PR #12, opened 2022-10-03
Nft as types

nmboavida:


bausano:
What is this change in relation to?

nmboavida:
> What is this change in relation to?

This change relates to the RFC standard:

- NFTs of a given NFT Collection have their own type T which is expressed in their Collection type Collection<T> , T being a witness type representing the NFT type

bausano (approved):
Perfect, Witness will be very useful

PR #17, opened 2022-10-10
Unified SupplyPolicy type

nmboavida:
Removed Collection Cap generic
Collection now has SupplyPolicy object


sources/collection/std_collection.move
```diff
@@ -46,7 +45,7 @@ module nft_protocol::std_collection {
         // Symbol of the Nft Collection. This parameter is a
         // vector of bytes that should enconde to utf8
         symbol: vector<u8>,
-        max_supply: Option<u64>,
+        max_supply: u64,
```
bausano, line 48: Unlimited is done with `u64::MAX`?

bausano: Nope, that's not the case upon reading the next changed LoC

sources/supply/policy.move
```diff
@@ -0,0 +1,70 @@
+//! Module contaning two supply `Cap` types, namely `Limited` and `Unlimited`.
```
bausano, line 1: Perhaps change this doc to refer to the `is_blind` flag.

Re: flag - not sure whether `is_blind` is a good name. Did you have some other ideas which you discarded in favour of `is_blind`?

nmboavida: I renamed it to `regulated` which inverts the logic. I made the necessary changes in all the conditions.

sources/collection/std_collection.move
nmboavida: To create a collection with unregulated supply the caller will set `max_supply = 0`. If `max_supply > 0` then the caller will be creating a collection with regulated supply. I added some documentation comments explaining the logic.

PR #19, opened 2022-10-10
Separate Collection from MintAuthority object

nmboavida:
Separate Collection from MintAuthority object


sources/collection/collection.move
```diff
@@ -58,16 +59,19 @@ module nft_protocol::collection {
         /// the royalty enforcement standard
         royalty_fee_bps: u64,
         creators: vector<Creator>,
+        /// The `Metadata` is a type exported by an upstream contract which is 
+        /// used to store additional information about the NFT.
+        metadata: M,
+    }
+
+    struct MintAuthority<phantom T> has key, store {
```
bausano, line 67: Perhaps there could be a `clone` function or something like that which enables multiple users be authorities?

PR #42, opened 2022-10-24
Toggling Sale Status Permissions

joaoaccaldeira:
The toggling of the sale status of a launchpad should be exclusive to its admin and not accessible to any person.


nmboavida:
This is indeed the intended behaviour, somehow I missed the assertion. Thank you! I will merge this asap.

PR #48, opened 2022-10-26
Safe + whitelisting

bausano:
Implementation of a proposal mentioned in this comment: https://github.com/MystenLabs/sui/pull/4887#issuecomment-1285537383


nmboavida:
> One implication of logical ownership:
> 
> In the case of embedded NFTs, they are currently being preminted to the Launchpad and therefore they initially exist in a state where they do not have logical owner until they get bought.
> 
> One solution is to change to flow as follows:
> 
> 1. Instead of NFT creators preminting the NFT to the launchpad, they will instead only create the `Data` object.
> 2. When the buyers purchase an NFT certificate, the launchpad will mint the NFT object initially as loose and then will call the function `nft::join_nft_data` to transform the loose NFT into an embedded NFT

In fact this means that the interaction surface can be significantly simplified as the embedded and loose NFTs will have pretty much the same flow from an interface standpoint

bausano:
TODO:
- emit events
- changelog
- docs
- some listing strategies such as blacklist or free-for-all

sources/safe/safe.move
```diff
@@ -0,0 +1,447 @@
...
+    use sui::vec_map::{Self, VecMap};
+    use std::type_name::{Self, TypeName};
+
+    struct Safe has key {
+        id: UID,
+        /// Accounting for deposited NFTs. Each NFT in the object bag is
+        /// represented in this map.
+        refs: VecMap<ID, NftRef>,
+        /// Holds the actual NFT objects as child objects.
+        nfts: ObjectBag,
```
Suficio, line 21: Should this still remain an `ObjectBag`? 
Something like a Table would be a bit simpler for the reasons we went with `Bag` for domains.

nmboavida (approved):
Fast-tracking to merge all to develop, let us add TODOs so we can tackle stuff in the next iterations

sources/safe/safe.move
bausano: https://github.com/Origin-Byte/nft-protocol/issues/94

PR #49, opened 2022-10-27
Dutch auction

Suficio:
Points that need to be resolved:
- Missing whitelist and multiple market endpoints
- Alternative to crit_bit?


sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+        reserve_price: u64,
+        /// A bid order stores amount of "T"okens the buyer is willing to
+        /// purchase 
+        bids: CBTree<vector<Bid>>,
+    }
+
+    /// An offer for a number of NFTs
+    struct Bid has store {
+        /// Total bid offer
+        offer: Balance<SUI>,
```
nmboavida, line 32: In general the word "offer" is  used interchangeable with "Ask" so I wonder if we can call this `amount` instead. I reckon the syrup orderbook is also using `order` so I will suggest to change it to `amount` there as well.

nmboavida:
TODO: We also need to add Gutenberg support for Dutch Auctions, ideally in this MR. I can walk you through Gutenberg in case it is helpful.

sources/launchpad/market/auction.move
bausano: OB uses `offer` 

```move
    /// An offer for a single NFT in a collection.
    struct Bid<phantom FT> has store {
        /// How many "T"okens are being offered by the order issuer for one NFT.
        offer: Balance<FT>,
        /// The address of the user who created this bid and who will receive an
        /// NFT in exchange for their tokens.
        owner: address,
        /// If the NFT is offered via a marketplace or a wallet, the
        /// faciliatator can optionally set how many tokens they want to claim
        /// on top of the offer.
        commission: Option<BidCommission<FT>>,
    }
```

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+    public entry fun create_bid<T>(
+        wallet: &mut Coin<SUI>,
+        slingshot: &mut Slingshot<T, Auction>,
+        tier_index: u64,
+        price: u64,
+        number: u64,
+        ctx: &mut TxContext,
+    ) {
+        // One can only place bids on NFT certificates if the slingshot is live
+        assert!(slingshot::live(slingshot) == true, err::launchpad_not_live());
```
bausano, line 94: alternatively

```
assert!(slingshot::live(slingshot), err::launchpad_not_live());
```

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+        /// purchase 
+        bids: CBTree<vector<Bid>>,
+    }
+
+    /// An offer for a number of NFTs
+    struct Bid has store {
+        /// Total bid offer
+        offer: Balance<SUI>,
+        /// The number of tokens for which the bid is placed
+        number: u64,
```
bausano, line 34: perhaps `quantity` if it refers to how many NFTs are being requested

However, I might be misunderstanding these data.

Would it simplify the logic if we had 1 bid per 1 NFT offer?

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+
+        if (crit_bit::has_key(&auction.bids, price)) {
+            vector::push_back(crit_bit::borrow_mut(&mut auction.bids, price), bid);
+        } else {
+            crit_bit::insert(&mut auction.bids, price, vector::singleton(bid));
+        }
+    }
+
+    /// Cancels a single order in a FIFO manner
+    // TODO: Potentially cancel all orders at price level?
```
bausano, line 250: I pondered this too in the OB. Perhaps a solution to this would be to merge `Bid`s of a user on the same price level? More complicated logic. The solution to introduce a function which cancels all is also good.

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+    }
+
+    fun refund_bid(bid: Bid, wallet: &mut Coin<SUI>, sender: &address) {
+        let Bid { offer, number: _number, owner } = bid;
+        assert!(sender == &owner, err::order_owner_must_be_sender());
+
+        balance::join(coin::balance_mut(wallet), offer);
+    }
+
+    fun conclude_auction<T>(
```
bausano, line 314: I suggest to create a new issue on GH to stress test this. For example, what if there are too many NFTs associated that we run out of compute units to actually conclude the auction. Perhaps we will need to have some pagination logic or something

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+        let fill_price = 0; 
+        let bids_to_fill = vector::empty();
+        while (nfts_to_sell > 0) {
+            if (crit_bit::is_empty(bids)) {
+                break
+            };
+
+            let max_key = crit_bit::max_key(bids);
+            // Do not allow price levels below reserve price to pass
+            if (max_key < reserve_price) {
```
bausano, line 335: Is it possible to prevent such insertion also in the `create_bid_` function?

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+                vector::destroy_empty(price_level);
+                continue
+            };
+
+            // There exists a bid we can match to an NFT
+            // Match in FIFO order
+            let bid = vector::remove(price_level, 0);
+
+            // Bid may exceed the number of NFTs left to sell
+            if (nfts_to_sell < bid.number) {
```
bausano, line 352: Perhaps this logic could be simplified if we required that each `Bid` is for exactly one NFT. On the other hand this approach is more versatile.

@nmboavida How could we address the issue where I could create a bid on a price level with like offer for 5 NFTs to snipe rare ones? I suppose without some randomness we cannot. The randomness, however, could also come from the call to `conclude_auction` which is done by the admin I belive.

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+
+            if (coin::value(&wallet) == 0) {
+                coin::destroy_zero(wallet);
+            } else {
+                transfer::transfer(wallet, owner);
+            }
+        };
+
+        vector::destroy_empty(bids_to_fill);
+    }
```
bausano, line 405: Could we also have something along the lines of `cancel_auction` just to refund the remaining bids which weren't filled? I suppose the admin can call `cancel_auction` now but I suggest we have a dedicated endpoint that perhaps does pretty much the same, but it's semantically different as one would expect that endpoint to be called after an auction is concluded.

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
+module nft_protocol::auction {
```
bausano, line 1: Let's discuss how can we make this module generic or how to duplicate its logic so that it can be used outside of launchpad as well

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,455 @@
...
+            price,
+            quantity,
+            tx_context::sender(ctx)
+        );
+
+        whitelist::burn_whitelist_token(whitelist_token);
+    }
+
+    /// Cancels a single bid at the given price level in a FIFO manner
+    fun cancel_bid<T>(
```
bausano, line 161: Should this be an `public entry`?

bausano (approved):
Some TODOs are left in the code, please either resolve them or create GH issues and link them there.

Other than that lgtm

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+        price: u64,
+        number: u64,
+        bidder: address,
+    ) {
+        let offer = balance::split(coin::balance_mut(wallet), price * number);
+
+        let bid = Bid { offer, number, owner: bidder };
+
+        if (crit_bit::has_key(&auction.bids, price)) {
+            vector::push_back(crit_bit::borrow_mut(&mut auction.bids, price), bid);
```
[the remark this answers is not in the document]
Suficio: Done

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+    // TODO: Potentially cancel all orders at price level?
+    // TODO: Cancel all bids enpoint
+    fun cancel_bid_(
+        book: &mut Auction,
+        wallet: &mut Coin<SUI>,
+        price: u64,
+        sender: address,
+    ) {
+        let bids = &mut book.bids;
+
```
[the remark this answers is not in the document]
Suficio: Done

sources/launchpad/market/auction.move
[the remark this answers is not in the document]
Suficio: Made bid refer to single NFT

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+            };
+
+            fill_price = max_key;
+            nfts_to_sell = nfts_to_sell - bid.number;
+            vector::push_back(&mut bids_to_fill, bid);
+        };
+
+        while (!vector::is_empty(&bids_to_fill)) {
+            let Bid {offer, number, owner} = vector::pop_back(&mut bids_to_fill);
+            let wallet = coin::from_balance(offer, ctx);
```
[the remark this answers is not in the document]
Suficio: Done

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+                    fill_price,
+                    receiver,
+                    ctx
+                );
+
+                let certificate = sale::issue_nft_certificate(sale, launchpad_id, ctx);
+
+                transfer::transfer(
+                    certificate,
+                    tx_context::sender(ctx),
```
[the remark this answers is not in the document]
Suficio: Done

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+            
+            while (number > 0) {
+                pay::split_and_transfer<SUI>(
+                    &mut wallet,
+                    fill_price,
+                    receiver,
+                    ctx
+                );
+
+                let certificate = sale::issue_nft_certificate(sale, launchpad_id, ctx);
```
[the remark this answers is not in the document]
Suficio: Done, checked for rest of file

sources/launchpad/market/auction.move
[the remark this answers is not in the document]
Suficio: Done

sources/launchpad/market/auction.move
```diff
@@ -0,0 +1,406 @@
...
+    use sui::tx_context::{Self, TxContext};
+    use sui::transfer;
+    use sui::object::{Self, UID, ID};
+
+    use nft_protocol::err;
+    use nft_protocol::slingshot::{Self, Slingshot};
+    use nft_protocol::crit_bit::{Self, CB as CBTree};
+    use nft_protocol::sale::{Self, Sale};
+
+    struct Auction has key, store {
```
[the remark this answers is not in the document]
Suficio: Done https://github.com/Origin-Byte/nft-protocol/issues/80

nmboavida (approved):

PR #55, opened 2022-11-02
Draft: Custom royalties

bausano:


sources/safe/royalties.move
```diff
@@ -0,0 +1,122 @@
...
+        payment: &mut TradePayment<C, W, FT>,
+    ): &mut Balance<FT> {
+        &mut payment.amount
+    }
+
+    public fun beneficiary<C, W, FT>(payment: &TradePayment<C, W, FT>): address {
+        payment.beneficiary
+    }
+
+    public fun transfer_remaining_to_beneficiary<C, W: drop, FT>(
```
Suficio, line 90: Shouldn't this function also drop the `TradePayment`?

bausano: Unfortunately, it is not yet possible to destruct shared objects. When this functionality is implemented, then yes.

bausano: https://github.com/Origin-Byte/nft-protocol/issues/95

Issue #54, opened 2022-11-01
Invalid module access. Unbound function 'transfer_to_object' in module '(sui=0x2)::transfer'


cgrewon:
https://github.com/Origin-Byte/nft-protocol/blob/main/sources/nft/nfts/unique_nft.move#L378

I have been faced error at this line, of course I am checking normal source, just tried to this command $sui move build

error message:
Invalid module access. Unbound function 'transfer_to_object' in module '(sui=0x2)::transfer'

$ sui move build
error[E03003]: unbound module member
    ┌─ ./sources/nft/nfts/unique_nft.move:378:9
    │
378 │         transfer::transfer_to_object(
    │         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Invalid module access. Unbound function 'transfer_to_object' in module '(sui=0x2)::transfer'

Failed to build Move modules: "Compilation error".

Sui version: $ sui --version
sui 0.13.2

Rustc version $ rustc --version
rustc 1.64.0 (a55dd71d5 2022-09-19)

PR #57, opened 2022-11-03
Upgrading to Sui 0.14.0

bausano:
### Changed

- Updated Sui dep to `0.14.0`
- Since `transfer_to_object` was deprecated, we now use `dynamic_object_field`
  with slingshot to associate embedded NFTs.


bausano:
Fixes #54

nmboavida (approved):

PR #85, opened 2022-11-21
New Protocol design

nmboavida:


bausano:
Let's use some of the domains and data models from this MR in a new one that is based off `develop`

PR #87, opened 2022-11-21
New protocol design #2: Refactor of custom royalties

bausano:


sources/safe/safe.move
```diff
@@ -202,37 +202,19 @@ module nft_protocol::safe {
         deposit_nft_(nft, safe, ctx);
     }
 
-    /// Remove an NFT from the `Safe` and give it back to the logical owner.
-    public entry fun withdraw_nft<T, D: store>(
```
Suficio, line 206: Why are you deleting this?

Suficio (approved):
One comment, otherwise everything lgtm!

sources/safe/safe.move
bausano: User can no longer freely take an NFT out of the safe in our latest proposal. Ie. NFTs always live in the Safe unless a creator whitelists a contract which enables taking it out.

PR #101, opened 2022-11-24
Refactor: Remove std collection

Suficio:


sources/collection/collection.move
```diff
@@ -68,9 +70,13 @@ module nft_protocol::collection {
...
         /// ID of `MintAuthority` object
         mint_authority: ID,
-        /// The `Metadata` is a type exported by an upstream contract which is
-        /// used to store additional information about the NFT.
-        metadata: M,
+        royalty_fee_bps: u64,
+        /// Domain storage equivalent to NFT domains which allows collections
+        /// to implement custom metadata.
+        ///
+        /// TODO: Should Collection domains be immutable?
```
bausano, line 77: Let's discuss the pros and cons. I wouldn't make it immutable, but an argument for append-only could be made.

bausano (approved):

PR #105, opened 2022-11-28
Feature: Royalty standard

Suficio:
Implements RoyaltyDomain for Collections and NFTs


sources/standards/royalty.move
```diff
@@ -0,0 +1,95 @@
+module nft_protocol::royalty {
+    use sui::tx_context::{Self, TxContext};
+
+    use nft_protocol::err;
+    use nft_protocol::nft::{Self, NFT};
+    use nft_protocol::collection::{Self, Collection};
+
+    struct RoyaltyDomain has store {
```
bausano, line 8: This is but one example of royalty collection. Perhaps we could name this something like `PercentageRoyaltyDomain` or `BpsRoyaltyDomain`. Cause and then there would be `ConstantRoyaltyDomain` for e.g. always 1 SUI, of whatever.

sources/standards/royalty.move
```diff
@@ -0,0 +1,95 @@
...
+    public fun receiver(domain: &RoyaltyDomain): address {
+        domain.receiver
+    }
+
+    public fun royalty_fee_bps(domain: &RoyaltyDomain): u64 {
+        domain.royalty_fee_bps
+    }
+
+    public fun calculate(domain: &RoyaltyDomain, amount: u64): u64  {
+        amount / 10000 * royalty_fee_bps(domain)
```
bausano, line 24: `10000` could be exported into a const

sources/standards/royalty.move
```diff
@@ -0,0 +1,95 @@
+module nft_protocol::royalty {
+    use sui::tx_context::{Self, TxContext};
+
+    use nft_protocol::err;
+    use nft_protocol::nft::{Self, NFT};
+    use nft_protocol::collection::{Self, Collection};
+
+    struct RoyaltyDomain has store {
+        /// Address that receives the mint and trade royalties
+        receiver: address,
```
bausano, line 10: Another royalty domain type mechanism could be such that the domain has a dynamic objects with relevant `Balance<FT>` into which these are stored. For now, let's defo go with this solution, just trying to point out that there might eventually exist multiple different royalty domains, so naming it more specifically is imho apt.

nmboavida (approved):

sources/standards/royalty.move
nmboavida: I added a TODO

sources/standards/royalty.move
nmboavida: I agree, I took the freedom to rename it, such that we can merge this PR.

bausano (approved):

PR #106, opened 2022-11-28
Feature: Tags standard

Suficio:


sources/standards/tags.move
```diff
@@ -0,0 +1,128 @@
...
+        bag::contains_with_type<Marker<T>, T>(&domain.bag, utils::marker<T>())
+    }
+
+    // TODO: Protect with AttributionDomain
+    public fun add_tag<T: store>(
+        domain: &mut TagDomain,
+        tag: T,
+        _ctx: &mut TxContext,
+    ) {
+        // TODO: Assert is creator
```
bausano, line 95: TODO

sources/standards/tags.move
```diff
@@ -0,0 +1,128 @@
...
+    // === Tags ===
+
+    struct Art has store, drop {}
+    struct PFP has store, drop {}
+    struct Collectible has store, drop {}
+    struct GameAsset has store, drop {}
+    struct TokenisedAsset has store, drop {}
+    struct Ticker has store, drop {}
+    struct DomainName has store, drop {}
+    struct Gif has store, drop {}
```
bausano, line 18: Not sure about this tag, seems like this could fit some metadata of the display domain rather?

sources/standards/tags.move
```diff
@@ -0,0 +1,128 @@
...
+    use sui::tx_context::TxContext;
+
+    use nft_protocol::utils::{Self, Marker};
+    use nft_protocol::nft::{Self, NFT};
+    use nft_protocol::collection::{Self, Collection};
+
+    // === Tags ===
+
+    struct Art has store, drop {}
+    struct PFP has store, drop {}
```
bausano, line 12: Can you add a comment what does PFP stand for please

sources/standards/tags.move
```diff
@@ -0,0 +1,128 @@
...
+    struct Collectible has store, drop {}
```
bausano, line 13: A note that a collectible is e.g. a baseball card, and main thing is that many instances of one type can exist (multiple players.)

sources/standards/tags.move
```diff
@@ -0,0 +1,128 @@
...
+    struct GameAsset has store, drop {}
+    struct TokenisedAsset has store, drop {}
```
bausano, line 15: What would be this?

sources/standards/tags.move
```diff
@@ -0,0 +1,128 @@
...
+    struct Ticker has store, drop {}
```
bausano, line 16: Not sure what to imagine under this tag

sources/standards/tags.move
nmboavida: I changed it to `ProfilePicture`

sources/standards/tags.move
nmboavida: A tokenised asset is a real life asset represented on-chain, such as an Insurance Policy, Loan contracts, etc.

sources/standards/tags.move
nmboavida: Tickers are what's called the abbreviation used to uniquely identify publicly traded assets, for instance Tesla trades as $TSLA and Amazon as $AMZN. The idea here is that asset tickers themselves can be minted and sold as NFTs.

nmboavida (approved):

PR #107, opened 2022-11-28
New Launchpad design

nmboavida:


sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+
+    use sui::object::{Self, ID , UID};
+    use sui::tx_context::{Self, TxContext};
+    use sui::object_table::{Self, ObjectTable};
+
+    use nft_protocol::err;
+    use nft_protocol::outlet::{Outlet};
+    use nft_protocol::generic::{Self, Generic};
+
+    struct Launchpad<phantom T> has key, store{
```
bausano, line 20:
```suggestion
    struct Launchpad<phantom T> has key, store {
```

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+        id: UID,
+        /// The address of the administrator
+        admin: address,
+        launches: ObjectTable<ID, Trebuchet>,
+    }
+
+    struct Trebuchet has key, store{
```
bausano, line 27:
```suggestion
    struct Trebuchet has key, store {
```

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+    use sui::tx_context::{Self, TxContext};
+    use sui::object_table::{Self, ObjectTable};
+
+    use nft_protocol::err;
+    use nft_protocol::outlet::{Outlet};
+    use nft_protocol::generic::{Self, Generic};
+
+    struct Launchpad<phantom T> has key, store{
+        id: UID,
+        /// The address of the administrator
```
bausano, line 22:
```suggestion
```

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    struct Trebuchet has key, store{
+        id: UID,
+        /// The ID of the Collections object
+        collections: vector<ID>,
+        /// Boolean indicating if the sale is live
+        live: bool,
+        /// The address of the administrator
+        admin: address,
```
bausano, line 34: There's already an admin field on the `Launchpad` itself. Can it happen that those two mismatch?

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+        id: UID,
+        /// The address of the administrator
+        admin: address,
+        launches: ObjectTable<ID, Trebuchet>,
+    }
+
+    struct Trebuchet has key, store{
+        id: UID,
+        /// The ID of the Collections object
+        collections: vector<ID>,
```
bausano, line 30: So one trebuchet can have many collections, and one launchpad many trebuchets then

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+    struct DeleteTrebuchetEvent has copy, drop {
+        object_id: ID,
+        collection_id: ID,
+    }
+
+    // === Launchpad Functions ===
+
+    /// Initialises a `Launchpad` object and adds it to the `Launchpad` object
+    public fun init_launchpad<T: drop>(
+        launchpad: &mut Launchpad<T>,
```
bausano, line 61: How do I instantiate `Launchpad` in the first place? Should we take this as a param, or should we just return a `Launchpad<T>` from this method?

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+            receiver,
+            markets,
+            is_embedded,
+            fee_config,
+        };
+
+        object_table::add(&mut launchpad.launches, id, slingshot);
+    }
+
+    // /// Burn the `Trebuchet`
```
bausano, line 118: TODO

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+
+    // === Modifier Functions ===
+
+    /// Toggle the Trebuchet's `live` to `true` therefore
+    /// making the NFT sale live.
+    public fun sale_on<T>(
+        _launchpad: &mut Launchpad<T>,
+        slingshot: &mut Trebuchet,
+    ) {
+        slingshot.live = true
```
bausano, line 151: you need to check that the lauchpad owns this trebuchet I believe

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /// Toggle the Trebuchet's `live` to `false` therefore
+    /// pausing or stopping the NFT sale.
+    public fun sale_off<T>(
+        _launchpad: &mut Launchpad<T>,
+        slingshot: &mut Trebuchet,
+    ) {
+        slingshot.live = false
```
bausano, line 160: you need to check that the lauchpad owns this trebuchet I believe

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /// Adds a sale outlet `Outlet` to `sales` field
+    public fun add_market<T>(
+        _launchpad: &mut Launchpad<T>,
+        slingshot: &mut Trebuchet,
+        market: Generic,
+    ) {
+        vector::push_back(&mut slingshot.markets, market);
```
bausano, line 169: you need to check that the lauchpad owns this trebuchet I believe

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    // === Getter Functions ===
+
+    /// Get the Trebuchet `id`
+    public fun id<T>(
```
bausano, line 175: `object::id(&slightshot)` works too, I'd remove this method

sources/launchpad/launchpad.move
nmboavida: That's right, however we need to reconsider how the flow of a sale with multiple collections in the next iterations

sources/launchpad/launchpad.move
nmboavida: Added TODO

sources/launchpad/launchpad.move
nmboavida: Done

sources/launchpad/launchpad.move
nmboavida: Done

sources/launchpad/launchpad.move
nmboavida: Updated to:

```
public entry fun init_launchpad(
        admin: address,
        receiver: address,
        auto_approval: bool,
        default_fee: ObjectBox,
        ctx: &mut TxContext,
    )
```

sources/launchpad/launchpad.move
nmboavida: Now there are essentially two admins, the Launchpad and the Slot admin

sources/utils/box.move
```diff
@@ -0,0 +1,78 @@
+module nft_protocol::box {
```
Suficio, line 1: If Box is unused then we should leave it out of the MR

Suficio (changes requested):

sources/utils/err.move
```diff
@@ -116,6 +120,30 @@ module nft_protocol::err {
...
+    public fun launchpad_slot_mismatch(): u64 {
+        return Prefix + 210
+    }
+
+    public fun wrong_admin(): u64 {
+        return Prefix + 212
+    }
+
+    public fun wrong_slot_admin(): u64 {
+        return Prefix + 212
```
bausano, line 132: Same error code as above, is that intentional?

sources/utils/err.move
```diff
@@ -190,4 +218,17 @@ module nft_protocol::err {
     public fun sender_not_owner(): u64 {
         return Prefix + 700
     }
+
+    // === Generic ===
```
bausano, line 222:
```suggestion
    // === Generic ===

```

sources/launchpad/whitelist.move
```diff
@@ -4,25 +4,25 @@ module nft_protocol::whitelist {
     use sui::tx_context::{Self, TxContext};
```
bausano, line 5: can we rename this module to `nft_protocol::launchpad_whitelist`, to make it easy to recognize from `nft_protocol::transfer_whitelist`. I'd keep the file name, just rename the module.

sources/launchpad/proceeds.move
```diff
@@ -0,0 +1,132 @@
...
+    use sui::transfer;
+    use sui::tx_context::TxContext;
+    use sui::object::{Self, UID};
+    use sui::balance::{Self, Balance};
+    use sui::dynamic_field as df;
+
+    struct Proceeds has key, store {
+        id: UID,
+        sold: Sold,
+        total: u64,
```
bausano, line 15: perhaps a comment - total of what?

bausano: does this total make sense tho? because we're adding basically different fungible tokens

like this would be sum of all tokens in all mints, is that a useful information?

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,434 @@
...
+        admin: address,
+        /// The address of the receiver of funds
+        receiver: address,
+        /// Vector of all markets outlets that, each outles holding IDs owned by the slot
+        markets: vector<ObjectBox>,
+        /// Field determining if NFTs are embedded or looose.
+        /// Embedded NFTs will be directly owned by the Slot whilst
+        /// loose NFTs will be minted on the fly under the authorithy of the
+        /// launchpad.
+        is_embedded: bool,
```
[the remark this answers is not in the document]
nmboavida: Removed

sources/launchpad/launchpad.move
```diff
@@ -0,0 +1,434 @@
...
+        slot: &mut Slot,
+        ctx: &mut TxContext,
+    ) {
+        assert_slot(launchpad, slot);
+        assert_launchpad_admin(launchpad, ctx);
+
+        slot.is_approved = true;
+    }
+
+    /// Adds a sale outlet `Outlet` to `sales` field
```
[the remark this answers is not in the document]
nmboavida: Resolved

sources/utils/box.move
nmboavida: Removed

sources/nft/nft.move
```diff
@@ -97,4 +97,18 @@ module nft_protocol::nft {
 
         nft.logical_owner = recipient;
     }
+
+    // === Getter Functions ===
+
+    public fun id<T>(
```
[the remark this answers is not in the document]
nmboavida: Removed

sources/utils/err.move
```diff
@@ -116,6 +120,30 @@ module nft_protocol::err {
         return Prefix + 210
     }
 
+    public fun launchpad_slot_mismatch(): u64 {
+        return Prefix + 210
+    }
+
+    public fun wrong_admin(): u64 {
```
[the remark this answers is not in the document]
nmboavida: I changed the name from wrong_admin() to `wrong_launchpad_or_slot_admin()`. This is in the case where the caller is either the slot or the launchpad admin, but errs if neither

sources/utils/err.move
nmboavida: It was a mistake, I fixed it now

sources/launchpad/proceeds.move
```diff
@@ -0,0 +1,132 @@
...
+    fun balance_mut<FT>(
+        proceeds: &mut Proceeds,
+    ): &mut Balance<FT> {
+        df::borrow_mut<TypeName, Balance<FT>>(
+            &mut proceeds.id,
+            type_name::get<Balance<FT>>(),
+        )
+    }
+
+    // public fun destroy_zero<FT>(
```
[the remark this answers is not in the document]
nmboavida: Added

examples/suimarines.move
```diff
@@ -60,19 +61,13 @@ module nft_protocol::suimarines {
         vector::push_back(&mut prices, 1000);
         vector::push_back(&mut prices, 2000);
 
-        fixed_price::create_market(
```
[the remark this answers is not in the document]
nmboavida: The flow is now the following:

1. Marketplace creates Launchpad
2. X creates a slot in the launchpad for an NFT Collection release (X can be the Marketplace or the NFT creator - if its the Marketplace the slot will be automatically approved, otherwise if its the NFT Creator creating the slot then it will only be automaticall approved if the flag `auto_approval = true`, otherwise the Marketplace has to approve it on their side)
3. Marketplace or NFT Creator can then add markets to the slot

In relation to collections, the launchpad markets are now agnostic to the collection type which means that multiple collections can be sold in the same market. We still need to work on this workflow a little bit more but the idea is that in the future you could sell NFTs from different collections in a bundle.

sources/launchpad/fees/fixed_fees.move
```diff
@@ -0,0 +1,67 @@
+module nft_protocol::flat_fee {
```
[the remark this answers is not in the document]
nmboavida: Makes a lot of sense, we can try to generalise fees/commissions/royalties, one thing to be cautious however is that in the case of the launchpad, fees can be charge in the bulk of the service instead of individual sales.

sources/launchpad/whitelist.move
nmboavida: Addressed

sources/launchpad/proceeds.move
nmboavida: My initial idea was that proceeds were all to be in the same fungible token and that therefore `Proceeds.total` would be the total of those proceeds. The reason for storing that info is to allow for custom types of fees that rely on the total FT-amount sold. Marketplaces could then reduce the amount of fees charged based on the bulk-volume of the sale. 

The same goes for `Sold` which is have renamed it to:

```
// Quantity of NFTs Sold
struct QtSold has copy, drop, store {
    // Bookeeping of all NFT sold whose proceeds have been withdrawn.
    collected: u64,
    // Total number of NFTs sold. The reason for storing this info is to
    // allow for custom types of fees that rely on the total FT-amount sold.
    // Marketplaces could then reduce the amount of fees charged based
    // on the bulk-volume of the sale.
    total: u64,
}
```


I have also added a TODO to reconsider `Proceeds.total` .

examples/suimarines.move
```diff
@@ -92,13 +87,38 @@ module nft_protocol::suimarines {
         royalties::transfer_remaining_to_beneficiary(Witness {}, payment, ctx);
     }
 
+    public entry fun mint_nft(
```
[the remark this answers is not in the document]
nmboavida: Removed

Suficio (approved):

PR #108, opened 2022-11-29
Feature: Attribution standard

Suficio:
Reimplements RoyaltyDomain to hold creator attributions.


sources/safe/transfer_whitelist.move
```diff
@@ -56,13 +58,13 @@ module nft_protocol::transfer_whitelist {
     /// collection to the whitelist, they can reexport this function in their
     /// module without the witness protection. However, we opt for witness
     /// collection to give the whitelist owner a way to combat spam.
-    public fun insert_collection<Admin: drop, T>(
+    public fun insert_collection<Admin: drop, T, FT>(
```
bausano, line 61: This is not great, let me think if we can remove the generic of FT somehow.

sources/standards/royalties/royalty.move
```diff
@@ -0,0 +1,142 @@
...
+    use sui::tx_context::TxContext;
+    use sui::dynamic_field::{Self as df};
+
+    use nft_protocol::attribution::{Self, Attributions, Creator};
+    use nft_protocol::royalty_strategy_bps::{Self, BpsRoyaltyStrategy};
+    use nft_protocol::royalty_strategy_constant::{
+        Self, ConstantRoyaltyStrategy
+    };
+
+    struct RoyaltyDomain<phantom FT> has key, store {
```
bausano, line 18: I'd remove the generic here and make the aggregator balance a dynamic object field.

sources/standards/royalties/attribution.move
```diff
@@ -194,4 +175,18 @@ module nft_protocol::attribution {
 
         assert!(bps_total == BPS, err::invalid_total_share_of_royalties());
     }
+
+    public fun assert_is_creator(
+        attributions: &Attributions,
+        who: address
+    ): u16 {
```
bausano, line 182: I suggest this returns nothing rather than the index.

Suficio: This is quite a useful convenience function which can be used to assert and withdraw the existence of creator.
Perhaps we could rename it?

sources/standards/royalties/royalty.move
```diff
@@ -68,57 +64,97 @@ module nft_protocol::royalty {
...
+        bag::remove<Marker<ConstantRoyaltyStrategy>, ConstantRoyaltyStrategy>(
+            &mut domain.strategies,
+            utils::marker<ConstantRoyaltyStrategy>()
+        );
     }
 
+    /// === Interoperability ===
+
     /// Calculate owed royalties from all the strategies defined on the domain
     public fun calculate(domain: &RoyaltyDomain, amount: u64): u64 {
```
bausano, line 132: Not so sure about utility of this function, let's discuss on a call

sources/standards/royalties/royalty.move
```diff
@@ -134,22 +170,30 @@ module nft_protocol::royalty {
...
-                b,
-            );
-        }
+        balance::join(aggregate, b);
     }
 
-    /// === Interoperability ===
+    /// === Utils ===
+
+    public fun assert_is_creator(domain: &RoyaltyDomain, who: address ) {
```
bausano, line 194:
```suggestion
    public fun assert_is_creator(domain: &RoyaltyDomain, who: address) {
```

sources/utils/utils.move
```diff
@@ -7,6 +7,13 @@ module nft_protocol::utils {
     use std::type_name;
     use std::vector;
 
+    /// Used to mark type fields in dynamic fields
+    struct Marker<phantom T> has copy, drop, store {}
+
+    public fun marker<T>(): Marker<T> {
```
bausano, line 13: Do you think we should expose this to the world, or keep it within our package only with `public(friend)`?

sources/standards/royalties/attribution.move
bausano: to get an index, it looks a bit odd:
```
let creator_index = assert_is_creator(&attrs, who);
```

to assert existence with a rename, depends on a name I suppose, can be also made look odd:
```
creator_index(&attrs, who);
```

I mean it's minor, but I reckon it's a bit better if we don't return an index. 

For that, perhaps:

```
public fun index(...): u16;
public fun try_index(...): Option<u16>;
```

Suficio: Yea lets do 
```
public fun index(...): u16;
public fun try_index(...): Option<u16>;
```

sources/standards/attribution.move
```diff
@@ -0,0 +1,196 @@
...
+            beneficiary.share_of_royalty_bps + creator.share_of_royalty_bps;
+    }
+
+    fun remove_creator(
+        attributions: &mut AttributionDomain,
+        ctx: &mut TxContext,
+    ): Creator {
+        let index = index(attributions, tx_context::sender(ctx));
+
+        vector::swap_remove(
```
bausano, line 169: This should come with an asterisk. We expose methods for getting index of a creator. However, using `remove_creator` invalidates the value of the last index and moves it.

I'd add this behavior to the docs for methods that remove creators, and work/return indexes.

sources/standards/royalties/royalty.move
```diff
@@ -0,0 +1,235 @@
+// TODO: Constant Royalty domain
+// TODO: Another royalty domain type mechanism could be such that the domain
```
bausano, line 2: TODO

sources/standards/royalties/royalty.move
```diff
@@ -0,0 +1,235 @@
+// TODO: Constant Royalty domain
```
bausano, line 1: TODO

sources/standards/attribution.move
bausano: this is the sort of thing that leads to exploits, perhaps it'd be better to use a hash map with key being the creator address?

sources/standards/royalties/royalty.move
```diff
@@ -0,0 +1,235 @@
...
+                &domain.strategies,
+                constant_strategy
+            );
+            royalty = royalty + royalty_strategy_constant::calculate(strategy);
+        };
+
+        royalty
+    }
+
+    public fun transfer_royalties<FT>(
```
bausano, line 169: If someone comes up with their own royalty policy, what would be the way they collect them? Ideally, I suggest we remove this function (and `calculate`) for now and let the creators go directly to the implementation if that's possible.

sources/standards/royalties/royalty.move
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    public fun royalty_domain_mut<C>(
+        nft: &mut NFT<C>,
+    ): &mut RoyaltyDomain {
+        nft::borrow_domain_mut(Witness {}, nft)
+    }
+
+    public fun collection_royalty_domain_mut<C>(
+        nft: &mut Collection<C>,
```
bausano, line 217: collection might end up being shared object. is it ok to give a mutable reference to royalty domain to anyone?

sources/standards/royalties/royalty.move
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    public fun collection_royalty_domain<C>(
+        nft: &Collection<C>,
+    ): &RoyaltyDomain {
+        collection::borrow_domain(nft)
+    }
+
+    public fun royalty_domain_mut<C>(
+        nft: &mut NFT<C>,
```
bausano, line 211: Not so clear to me just yet what's the use case for a royalty domain being on NFTs. Do you have some ideas around this? If not yet, I suggest we make an issue to discuss it at a later point and remove all functionality that does royalty for NFTs and retain only for collection.

sources/standards/royalties/royalty_strategy.move
```diff
@@ -0,0 +1,39 @@
...
+        /// Royalty charged on trades in basis points
+        royalty_fee_bps: u64,
+    }
+
+    public fun royalty_fee_bps(domain: &BpsRoyaltyStrategy): u64 {
+        domain.royalty_fee_bps
+    }
+
+    public fun calculate(domain: &BpsRoyaltyStrategy, amount: u64): u64  {
+        amount / BPS * royalty_fee_bps(domain)
```
bausano, line 14: https://github.com/MystenLabs/sui/blob/main/crates/sui-framework/deps/move-stdlib/sources/fixed_point32.move

```
fixed_point32::multiply_u64(
  fixed_point32::create_from_rational(amount, BPS),
  royalty_fee_bps(domain),
)
```

to minimise inaccuracy

sources/standards/attribution.move
Suficio: Done

sources/standards/royalties/royalty.move
Suficio: Deleted

sources/standards/royalties/royalty.move
Suficio: Yes, all mutable access functions into `RoyaltyDomain` itself are protected by `AttributionDomain`.

sources/standards/royalties/royalty.move
Suficio: This just leaves the possibility open for users to create per-NFT policies.

Removed for now.

sources/standards/royalties/royalty_strategy.move
Suficio: Done

sources/standards/attribution.move
```diff
@@ -0,0 +1,201 @@
...
+    }
+
+    public fun share_of_royalty_bps(creator: &Creator): u16 {
+        creator.share_of_royalty_bps
+    }
+
+    struct AttributionDomain has store {
+        /// Address that receives the mint and trade royalties
+        creators: VecMap<address, Creator>,
+    }
```
nmboavida, line 39: Maybe we can have only one data structure `AttributionDomain` as:

```
struct AttributionDomain has store {
    creators: VecMap<address, u64>,
    total_share: u64,
}
```

Where `u64` in `VecMap<address, u64>` is the share of royalty, and 
`total_share` is the sum of all shares.

bausano (approved):

sources/standards/attribution.move
nmboavida: Alternatively we could have:

```
struct AttributionDomain has store {
    creators: VecMap<address, Share>,
    total_share: u64,
}

struct Share has store {
    share: VecMap<address, Share>,
    dilutable: bool,
}
```

The idea is to emulate how shares work in the real world. Usually the project founders will hold dilutable equity and will dilute their share, but the shares given may not be dilutable thereafter. This emulates how the logic in `add_creator` however this scheme allows us to dilute the shares accross all dilutable royalty shareholders, instead of only the transaction sender.

The challenge here is that at least 51% of dilutable shareholders should vote to add a new creator

nmboavida (approved):

PR #115, opened 2022-11-30
Feature: Flyweight domain

nmboavida:


sources/standards/class.move
```diff
@@ -0,0 +1,136 @@
+// TODO: Where does it make sense to control supply?
+// TODO: Where does it make sense to control ownership of the shared data object?
+// This should ideally be controlled solely by the NFT Creators..
+module nft_protocol::class {
```
bausano, line 4: Is there a better name? It's quite generic, perhaps that was a point, but it's hard to sort of imagine what `class` might do. It seems to behave similarly to NFT or collection

sources/utils/err.move
```diff
@@ -40,6 +40,10 @@ module nft_protocol::err {
         return Prefix + 09
     }
 
+    public fun mint_authority_mistmatch(): u64 {
```
bausano, line 43:
```suggestion
    public fun mint_authority_mismatch(): u64 {
```

sources/standards/class.move
```diff
@@ -0,0 +1,136 @@
...
+
+    public fun add_domain<C, V: store>(
+        class: &mut Class,
+        mint: &MintAuthority<C>,
+        v: V,
+        ctx: &mut TxContext,
+    ) {
+        assert!(
+            collection::mint_id(mint) == class.mint_authority,
+            err::mint_authority_mistmatch()
```
bausano, line 109:
```suggestion
            err::mint_authority_mismatch()
```

Suficio:
In general, I wanted to propose that we merge Collection into NFT and implement any logic using domains.

This introduces a third domain-based abstraction where the logic doesn't cover much than Collection already does leading to a bunch of duplicated code. With a couple more PRs adding domains to Collections we should be able to get rid of it entirely and replace it with an NFT with a MintAuthority domain.

sources/standards/class.move
nmboavida: You're right, the intent is to behave like an NFT from a data perspective but to be shared - to be the data side of a loose implementation. Do you have any name recommendations?

bausano: https://refactoring.guru/design-patterns/flyweight

sources/standards/class.move
```diff
@@ -0,0 +1,136 @@
+// TODO: Where does it make sense to control supply?
```
bausano, line 1: Done

bausano (approved):
There are some TODOs which we might want to convert into issues. Also there's Kuba's discussion to be resolved.

```
// TODO: Where does it make sense to control supply?
// TODO: Where does it make sense to control ownership of the shared data object?
// This should ideally be controlled solely by the NFT Creators..
```

sources/collection/collection.move
```diff
@@ -163,7 +163,7 @@ module nft_protocol::collection {
 
     public fun add_domain<C, V: store>(
         collection: &mut Collection<C>,
-        mint_cap: &mut MintCap<C>,
+        mint_cap: &MintCap<C>,
```
Suficio, line 166: `&mut MintCap<C>` is necessary to prove ownership of `MintCap`

sources/standards/flyweight.move
```diff
@@ -0,0 +1,129 @@
...
+
+    // === Domain Functions ===
+
+    public fun borrow_nft<C>(state: &Archetype<C>): &Nft<C> {
+        &state.nft
+    }
+
+    public fun borrow_nft_mut<C>(
+        state: &mut Archetype<C>,
+        _mint: &MintCap<C>,
```
Suficio, line 125: `&mut MintCap<C>` is necessary to prove ownership of MintCap

sources/standards/flyweight.move
```diff
@@ -0,0 +1,129 @@
...
+
+    struct Pointer has key, store {
+        id: UID,
+        data: ID,
+    }
+
+    struct Archetype<phantom C> has key, store {
+        id: UID,
+        nft: Nft<C>,
+        supply: Supply,
```
Suficio, line 22: Why supply?

sources/standards/flyweight.move
```diff
@@ -0,0 +1,129 @@
...
+        mint_authority: ID,
```
Suficio, line 23: Mutable access to `Archetype` should be governed by `AttributionDomain` as is in other domains

sources/standards/flyweight.move
```diff
@@ -0,0 +1,129 @@
...
+    use nft_protocol::collection::{Self, Collection, MintCap};
+
+    struct Pointer has key, store {
+        id: UID,
+        data: ID,
+    }
+
+    struct Archetype<phantom C> has key, store {
+        id: UID,
+        nft: Nft<C>,
```
Suficio, line 21: The existence of `logical_owner` on NFT makes this pattern impossible as transactions adding domains must be signed by the `logical_owner`.

sources/collection/collection.move
nmboavida: Imo using &mut as an ownership test is a dangerous pattern and I would try to avoid it. In any case, if an object is private, then &Object can only be used in a function call by its owner

sources/standards/flyweight.move
```diff
@@ -0,0 +1,129 @@
...
+    use nft_protocol::nft::{Self, Nft};
+    use nft_protocol::supply::{Self, Supply};
+    use nft_protocol::collection::{Self, Collection, MintCap};
+
+    struct Pointer has key, store {
+        id: UID,
+        data: ID,
+    }
+
+    struct Archetype<phantom C> has key, store {
```
Suficio, line 19: Lets call it `ArchetypeDomain` to keep pattern

sources/standards/flyweight.move
```diff
@@ -0,0 +1,129 @@
...
+    use sui::event;
+    use sui::object::{Self, UID, ID};
+    use sui::tx_context::{TxContext};
+    use sui::object_table::{Self, ObjectTable};
+
+    use nft_protocol::nft::{Self, Nft};
+    use nft_protocol::supply::{Self, Supply};
+    use nft_protocol::collection::{Self, Collection, MintCap};
+
+    struct Pointer has key, store {
```
Suficio, line 14: Pointer should store `collection_id` and `archetype_id`, as you wont be able to directly point to a dynamic field of an object.

PR #127, opened 2022-12-06
Feature: Mutable API for Collection domains 

Suficio:
- Cleans up Collection a bit
- Mutable API for display standards
- Mutable API for TagDomain
- Interoperability API for AttributionDomain

### DIFF
--- a/examples/suimarines.move
+++ b/examples/suimarines.move
@@ -26,7 +26,6 @@ module nft_protocol::suimarines {
         let (mint_cap, collection) = collection::create<SUIMARINES>(
             &witness,
             100, // max supply
-            false, // is mutable
             ctx,
         );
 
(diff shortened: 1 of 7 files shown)


nmboavida:
I think it would be great to have a domain to define if the collection is frozen or not. If the collection is frozen then nothing can be added or changed to any collection domain. The same can be made to freeze the NFTs. This is sometimes desired to prevent malicious creators from altering the collection post-mint (i.e. rugpulls).

nmboavida (changes requested):
I think it would be great to have a domain to define if the collection is frozen or not. If the collection is frozen then nothing can be added or changed to any collection domain. The same can be made to freeze the NFTs. This is sometimes desired to prevent malicious creators from altering the collection post-mint (i.e. rugpulls).

In relation to mutability of NFTs I am in favour of having these permissioning systems defined on the collection level in order to avoid data redundancy, but let me know if you have any other suggestions.

Let us consider this for a future iteration.

PR #139, opened 2022-12-12
Fix: Rename NFT to Nft

nmboavida:
Struct names in capital letters, equal to the name of their module, are reserved to [One-Time Witnesses](https://examples.sui.io/basics/one-time-witness.html).


Suficio (approved):
LGTM, did @bausano not have some comments against this in the past?

PR #142, opened 2022-12-12
Feature: Attributes domain

nmboavida:


sources/standards/display.move
```diff
@@ -241,4 +241,48 @@ module nft_protocol::display {
     ) {
         collection::add_domain(nft, mint_cap, new_symbol_domain(symbol));
     }
+
+    /// === AttributesDomain ===
+
+    struct Attributes has store {
+        keys: vector<String>,
```
Suficio, line 248: Is this better solved using a `VecMap`, as we would have to check whether keys and values have the same length otherwise?

Would also be good to add some method obtain a mutable reference to keys and values.

sources/standards/display.move
```diff
@@ -241,4 +241,48 @@ module nft_protocol::display {
...
+
+    public fun display_attribute_keys<C>(nft: &NFT<C>): &vector<String> {
+        keys(nft::borrow_domain<C, Attributes>(nft))
+    }
+
+    public fun display_attribute_values<C>(nft: &NFT<C>): &vector<String> {
+        values(nft::borrow_domain<C, Attributes>(nft))
+    }
+
+    public fun add_attributes_domain<C>(
```
Suficio, line 280: Think we should modify all `add_domain` functions to take a `&mut Collection<C>` and require that `ctx` is a creator from `AttributionDomain`, otherwise anyone could add this domain if creators choose not to use it.

Suficio (changes requested):

sources/standards/display.move
nmboavida: Changed attributes to:

```
struct Attributes has store {
    map: VecMap<String, String>,
}
```


and added:

```
public fun display_attribute_mut<C>(nft: &mut Nft<C>): &mut Attributes {
    nft::borrow_domain_mut<C, Attributes, Witness>(Witness {}, nft)
}
```

sources/standards/display.move
nmboavida: Addressed.

Suficio (approved):

PR #144, opened 2022-12-12
Fix: Use indirect market reference

Suficio:
Cannot get mutable reference to dynamic fields of object


nmboavida:
Alternatively we could have a more restrictive permissioned system:

```
friend nft_protocol::fixed_price;
friend nft_protocol::dutch_auction;

public(friend) fun market_internal_mut() {}
```

nmboavida (approved):

PR #151, opened 2022-12-15
Tests: NFT domains

nmboavida:


sources/nft/nft.move
```diff
@@ -96,4 +96,13 @@ module nft_protocol::nft {
 
         nft.logical_owner = recipient;
     }
+
+    // === Test only helpers ===
+
+    #[test_only]
```
bausano, line 102: I reckon we can export this as a non test getter too

tests/domains/display.move
```diff
@@ -0,0 +1,197 @@
...
+        let mint_cap = test_scenario::take_from_address_by_id<MintCap<Foo>>(
+            &scenario, CREATOR, cap_id
+        );
+
+        display::add_collection_display_domain(
+            &mut collection,
+            &mint_cap,
+            string::utf8(b"Suimarines-234"),
+            string::utf8(b"Collection of Suimarines"),
+        );
```
bausano, line 66: perhaps another statement that attempts to `get_domain<DisplayDomain>` so that if this method was a no-op the test case would fail

bausano (approved):

tests/nft.move
```diff
@@ -0,0 +1,149 @@
...
+        // If domain does not exist this function call will fail
+        nft::borrow_domain<Foo, DomainA>(&nft);
+
+        transfer(nft, OWNER);
+
+        test_scenario::end(scenario);
+    }
+
+    // #[test]
+    // fun borrows_domain_mut() {
```
Suficio, line 59: Do we fix this now?

Suficio (changes requested):

tests/nft.move
```diff
@@ -0,0 +1,149 @@
...
+        let scenario = test_scenario::begin(OWNER);
+        let ctx = ctx(&mut scenario);
+
+
+        let nft = nft::new<Foo>(OWNER, ctx);
+
+        nft::add_domain(&mut nft, DomainA {}, ctx);
+
+        nft::borrow_domain<Foo, DomainA>(&nft);
+
```
Suficio, line 140: Double spacing

Suficio (changes requested):

sources/nft/nft.move
nmboavida: Done

tests/nft.move
nmboavida: Done

tests/domains/display.move
nmboavida: Oh that's right, I just added it to all tests, thanks!

Suficio (approved):

PR #152, opened 2022-12-16
Tests: Bidding with royalties

nmboavida:


sources/standards/royalties/royalty.move
```diff
@@ -115,6 +115,8 @@ module nft_protocol::royalty {
         amount: u64,
     ): u64 {
         if (!contains_royalty_strategy<BpsRoyaltyStrategy>(domain)) {
+            // TODO: This is dangerous because it can lead to
```
bausano, line 118: can we create a new issue with some proposed solutions and _why_ it's dangerous

sources/standards/royalties/royalty_strategy.move
```diff
@@ -11,7 +11,9 @@ module nft_protocol::royalty_strategy_bps {
     }
 
     public fun calculate(domain: &BpsRoyaltyStrategy, amount: u64): u64  {
-        amount / BPS * royalty_fee_bps(domain)
+        // TODO: Need to implement Decimals module for calculation
```
bausano, line 14: Sam has just implemented something like this for Move, we can wait for that MR to land.

sources/trading/bidding.move
```diff
@@ -161,9 +163,13 @@ module nft_protocol::bidding {
 
         let nft_id = safe::transfer_cap_nft(&transfer_cap);
 
+        // TODO: This function name is confusing, since the
```
bausano, line 166: This can be called by anyone I believe, perhaps `settle_funds`?

sources/trading/bidding.move
```diff
@@ -161,9 +163,13 @@ module nft_protocol::bidding {
...
+        // caller of the function is the seller and not the buyer.
+        // Shall we rename it to receive_funds?
         pay_for_nft<C, FT>(
             &mut bid.offer,
-            bid.buyer,
+            tx_context::sender(ctx),
+            // bid.buyer,
```
bausano, line 172: I need to validate that this is ok

bausano: This needs to be the buyer, not the tx sender which would be the seller in this case.

bausano (approved):

sources/safe/royalties.move
```diff
@@ -120,4 +120,12 @@ module nft_protocol::royalties {
...
         });
     }
+
+    // === Getters ===
+
+    public fun amount_u64<C, FT>(
+        trade_payment: &TradePayment<C, FT>,
+    ): u64 {
+        balance::value(&trade_payment.amount)
+    }
```
Suficio, line 130: People can call `balance::value` on the returned `Balance`

```suggestion
    public fun amount<C, FT>(
        trade_payment: &TradePayment<C, FT>,
    ): &Balance<FT> {
        &trade_payment.amount
    }
```

sources/standards/royalties/royalty.move
Suficio: IMO if the royalty isn't defined then its contribution to the total royalty owed is zero

sources/standards/royalties/royalty_strategy.move
Suficio: There are some snippets of code that use floats for this calculation in our codebase already.

This definetly shouldn't be precedent to implement a decimal library as we're dealing with fixed point arithmetic.

Suficio (changes requested):

sources/standards/royalties/royalty.move
nmboavida: The problem is if the user thinks it has defined the royalties and but it has not, this function will still run, whilst imo it shoudln't

tests/utils_2.move
```diff
@@ -0,0 +1,130 @@
+// TODO: This module is to be merged with test_utils from
```
bausano, line 1: todo

sources/standards/royalties/royalty.move
nmboavida: Issue: https://github.com/Origin-Byte/nft-protocol/issues/158

sources/trading/bidding.move
nmboavida: I agree, I renamed it to `settle_funds`

Suficio (approved):

PR #157, opened 2022-12-19
Feature: Single Inventory per Slot

Suficio:
Multiple Inventories per Slot overcomplicates the model while only different primary markets per FT are a useful business case.

### DIFF
--- a/examples/suimarines.move
+++ b/examples/suimarines.move
@@ -3,7 +3,6 @@ module nft_protocol::suimarines {
     use std::string;
 
     use sui::balance;
-    use sui::object::ID;
     use sui::transfer::transfer;
     use sui::tx_context::{Self, TxContext};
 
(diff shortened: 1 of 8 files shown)


bausano:
https://github.com/Origin-Byte/nft-protocol/pull/170#issuecomment-1361177149 It does have some benefits though. Moving our discussions here:

> @Suficio: The idea I had was that Inventory is append only, and can be single-writer during the minting process

> @nmboavida exactly, which is what this MR is doing, however it still needs &mut Inventory , and therefore even though it is a SW-object each transaction puts a lock on the objects while being processed

> @bausano Unless you reference it with &Inventory it for sure will not be possible to paralyze it.
If you reference it with &Inventory, there shouldn't be anything in theory to prevent parallelization. However, depends on the Sui implementation here really.

> Additionally, we should note to the devs, that they should split their Coin gas into many smaller objects. Because that's another object on which parallelization will fail if not taken into account

> @Suficio After finishing minting, the Inventories should be transfered to the Market

Suficio:
You are still able to create multiple inventories and merge them into the market.

Suficio:
The big advantage of this approach is simpler logic but it also allows users to implement something like an auction with an instant buy, or fixed price purchases for the same inventory in different FT.

However, an alternative approach is defining markets on the Inventory directly instead of the Slot.

Suficio:
Replaced by #178

PR #161, opened 2022-12-19
Publish script automatically gets a gas object from CLI

bausano:


Suficio (approved):
I think this happens automatically, when i publish I don't provide the gas object.

Could be easier to just do an if `$GAS` is set, pass the argument?

bausano:
Ah I see, you're right. Changed the script

PR #164, opened 2022-12-20
Docs and improvements for Orderbook

bausano:
### Added

- Creating an ask with commission in the orderbook returns an error if the
  commission is gtq to the ask price.

### Changed

- Renamed some arguments in the orderbook to be more descriptive.

### Removed

- Some protected actions were entry methods. This did not make sense because
  witness shouldn't have the `key` ability and therefore cannot be an entry
  function argument.


CHANGELOG.md
```diff
@@ -6,6 +6,22 @@ The format is based on [Keep a
 Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.1.0] - 2022-12-19
+
+### Added
+
+- Creating an ask with commission in the orderbook returns an error if the
+  commission is gtq to the ask price.
```
Suficio, line 14: Remove acronyms

docs/ORDERBOOK.md
```diff
@@ -0,0 +1,264 @@
...
+- instantly buy a specific NFT
+
+# Intermediary state
+
+By working with `Safe`, clients must know _up front_ which exact `Safe`
+instance to provide into the entry methods for transferring an NFT.
+However, OB concept supports frequent trades with `create_ask` and `create_bid`
+endpoints.
+In frequent trades, the NFT changes fast and therefore the `Safe` instance
+cannot reliably be known in advance.
```
Suficio, line 29:
```suggestion
When trading frequently, the `Safe` in which an NFT lives often cannot be known
in advance. This becomes a problem as clients must know _up front_ which exact
`Safe` instance to provide into the `create_ask` and `create_bid` orderbook
endpoints.

```

docs/ORDERBOOK.md
```diff
@@ -0,0 +1,264 @@
...
+reward for the NFT.
+For example, if an NFT is sold for 10 SUI and commission was 2 SUI, then the
+NFT seller receives 8 SUI and the beneficiary receives 2 SUI.
+Once the ask is matched, then the commission funds are _eventually_ transferred
+to the beneficiary address.
+See the [documentation](#intermediary-state) for the intermediary state above.
+The commission is paid to the beneficiary only after the intermediary state
+have been resolved.
+
+To receive a commission on an instant NFT buy (when buying a specific NFT)
```
Suficio, line 69: Unclear what you mean

docs/ORDERBOOK.md
```diff
@@ -0,0 +1,264 @@
...
+  beneficiary: address,
+  commission_ft: u64,
+  wallet,
+)
+```
+
+To cancel an existing bid, the client gives the price they sent as an input in
+the previous endpoint.
+If there are multiple bids the tx sender has created with the same price, then
+only one is cancelled.
```
Suficio, line 167:
```suggestion
only the first order is cancelled.
```

Suficio (approved):

docs/ORDERBOOK.md
bausano: Let me reword it, but 

> This becomes a problem as clients must know _up front_ which exact `Safe` instance to provide into the `create_ask` and `create_bid` orderbook

Is not correct. They must know if only for transfer. In fact, it is the case that for those two endpoints they don't need to know it by design.

CHANGELOG.md
bausano: done

docs/ORDERBOOK.md
Suficio: > This becomes a problem as clients would have to know up front which exact Safe instance to provide into the create_ask and create_bid orderbook

Perhaps would be better. What was unclear to me was whether you were describing the hypothetical problem situation or the solution, so I suggested to rewrite describing the problem situation which would then be followed by the solution.

docs/ORDERBOOK.md
bausano: reworded

docs/ORDERBOOK.md
```diff
@@ -0,0 +1,274 @@
...
+reward for the NFT.
+For example, if an NFT is sold for 10 SUI and commission was 2 SUI, then the
+NFT seller receives 8 SUI and the beneficiary receives 2 SUI.
+Once the ask is matched, then the commission funds are _eventually_ transferred
+to the beneficiary address.
+See the [documentation](#intermediary-state) for the intermediary state above.
+The commission is paid to the beneficiary only after the intermediary state
+have been resolved.
+
+If a marketplace or a wallet want to receive a commission on a trade they
```
Suficio, line 76:
```suggestion
If a marketplace or a wallet wants to receive a commission on a trade they
```

Suficio (approved):

PR #170, opened 2022-12-20
Feature: Fast Mints

nmboavida:
The minting pattern is now:

1. Create a private Inventory
2. Mass mint NFTs to this private Inventory
4. Create a Market and transfer the Private Inventory

To simplify, I have decided to remove `mint_cap.supply_policy`  and it the creators want they can create a domain on the Collection to keep track of the supply.

As a consequence of this new design, NFTs are now owned on the Invetory level, rather than on the Slot level, simply because the Slot is a shared object all the time, whilst Inventory can be instantiated as a private object and subsequently transferred to the Market, which is owned by the Slot.


examples/suimarines.move
```diff
@@ -123,6 +118,6 @@ module nft_protocol::suimarines {
             ctx,
         );
 
-        slot::add_nft(slot, market_id, nft, ctx);
+        inventory::add_nft(inventory, nft);
```
Suficio, line 121: Examples should also create a `Slot` and insert the `Inventory` into it

sources/launchpad/inventory.move
```diff
@@ -31,26 +36,38 @@ module nft_protocol::inventory {
         // NFTs that are currently on sale. When a `NftCertificate` is sold,
         // its corresponding NFT ID will be flushed from `nfts` and will be
         // added to `queue`.
-        nfts: vector<ID>,
+        nfts_on_sale: vector<ID>,
         // NFTs whose certificates have been sold and currently waiting
         // to be redeemed. When a `NftCertificate` is redeemed, its respective
         // NFT ID is flushed out of `queue`
+        // TODO: We can most likely deprecate this queue in favour of simply
```
Suficio, line 43: We aren't even using it right now, might as well remove as part of this MR

Suficio (approved):
Please add a unit test that demonstrates the process of inserting an `Inventory` to `Slot`.

The function to init them, `init_slot` is already in `tests/slot`.

bausano:
I changed the base from `main` to `develop`, conflicts

sources/launchpad/inventory.move
```diff
@@ -31,26 +36,38 @@ module nft_protocol::inventory {
...
+        // having the NFTs in dynamic fields - that itself is already performing the
+        // accounting
         queue: vector<ID>,
     }
 
+    public entry fun create_and_transfer(
```
bausano, line 49: minor: we usually use `create_for_sender`

sources/launchpad/inventory.move
```diff
@@ -71,29 +88,63 @@ module nft_protocol::inventory {
...
             queue: _,
         } = inventory;
 
         object::delete(id);
     }
 
+    /// Adds NFT as a dynamic child object with its ID as key and
+    /// adds an NFT's ID to the `nfts` field in `Inventory` object.
+    ///
+    /// This should only be callable when Inventory is private and not
```
bausano, line 101: I wonder whether this is testable with the `test_scenario` module

sources/launchpad/inventory.move
```diff
@@ -71,29 +88,63 @@ module nft_protocol::inventory {
...
+    /// owned by the Slot. The function call will fail otherwise, because
+    /// one would have to refer to the Slot, the parent shared object, in order
+    /// for the bytecode verifier not to fail.
+    public entry fun add_nft<C>(
+        inventory: &mut Inventory,
+        nft: Nft<C>,
+    ) {
+
```
bausano, line 109:
```suggestion
```

bausano:
This MR is a step in the right direction for mass-minting NFTs. However, this problem is not yet solved if I understand it correctly. We're still dependent on `&mut Invetory`.

Can we use several inventories with launchpad? Then we could create let's say 1 inventory per 100 NFTs. So if someone is creating 10k NFTs, they'd create 100 inventories, and parallelize the insertion to those.

examples/suimarines.move
nmboavida: Added in test module `mint_and_sell`

sources/launchpad/inventory.move
nmboavida: I will keep it for the timebeing because we are using it in the delete function, and given that we will most likely restructure how inventory work this field will inevitably be removed.

PR #177, opened 2022-12-27
Feature: Remove NFT Certificate

nmboavida:
The intial purpose of NftCertificate was to allow the Launchpad to emit a certificate with an NFT ID, such that the user could then use this Certificate in a second function call to redeem the NFT with that ID. This had to be done because there were no Dynamic Fields before. Now, with the introduction of dynamic fields this two step process can be simplified to one step, and as a result, it improves the UX of the Marketplaces.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 sui.log.*
 .env
 build
+target
 output.bpl
 .coverage_*
 .trace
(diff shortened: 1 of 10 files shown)


nmboavida:
The intial purpose of NftCertificate was to allow the Launchpad to emit a certificate with an NFT ID, such that the user could then use this Certificate in a second function call to redeem the NFT with that ID. This had to be done because there were no Dynamic Fields before. Now, with the introduction of dynamic fields this two step process can be simplified to one step, and as a result, it improves the UX of the Marketplaces.

Suficio (approved):
Remove commented lines please

PR #178, opened 2022-12-27
Feature: Register markets on Inventory

Suficio:
Registers Markets on Inventory allowing users to sell one Inventory through different sales channels.

### DIFF
--- a/examples/suitraders.move
+++ b/examples/suitraders.move
@@ -81,16 +81,22 @@ module nft_protocol::suitraders {
             ctx,
         );
 
-        nft_protocol::fixed_price::init_market<sui::sui::SUI>(
+        let inventory_id =
+            nft_protocol::slot::create_inventory(&mut slot, false, ctx);
+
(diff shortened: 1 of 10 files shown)


examples/suitraders.move
```diff
@@ -81,16 +81,22 @@ module nft_protocol::suitraders {
             ctx,
         );
 
-        nft_protocol::fixed_price::init_market<sui::sui::SUI>(
+        let inventory_id =
+            nft_protocol::slot::create_inventory(&mut slot, false, ctx);
```
bausano, line 85: I wish we had enum types so that we don't have to use bools.

I think it'd make this example more readable if you assigned the `false` to an aptly named variable and used the variable as the 2nd arg.

nmboavida (approved):

nmboavida:
Lets add Changelog not before merging

nmboavida:
Lets add changelog note before merging.

PR #184, opened 2022-12-30
Plugin pattern

bausano:
### Added

- `PluginDomain` which collects type names of witness structs.
  These then serve to authorize a function in a "base" contract (the one that is deployed with the one-time-witness type.)
  The function returns the original witness type (of the same module as the OTW) in exchange for a witness of another smart contract (a plugin) if that plugin's witness is present in the `PluginDomain`.
- `Multisig` is a utility struct which enables smart contracts to authorize
  actions signed by predefined set of accounts.

### Changed

- Adding a creator now accepts a mutable reference to the collection instead of
  the `CreatorDomain`.
  This is because getting the reference to the `CreatorDomain` required a `&mut TxContext`.
  The same reference was required in the `add_creator` function.


sources/utils/multisig.move
```diff
@@ -0,0 +1,154 @@
...
+    use std::option::{Self, Option};
+
+    use sui::vec_map::{Self, VecMap};
+    use sui::object::{Self, UID};
+    use sui::tx_context::{Self, TxContext};
+
+    use nft_protocol::attribution;
+    use nft_protocol::collection::Collection;
+
+    struct Multisig<T> has key, store {
```
Suficio, line 14: `MultiSig` should be constructible from a list of addresses and droppable only once those addresses have been removed to make it maximally re-usable.

Can you think of examples where aggregating the bps share of signers would be useful?
We assumed in other domains that being a creator grants you full administrative power. We could even consider separating the bps share which was used for royalty calculations into the RoyaltyDomain.

Its a good candidate for adding to Originmate.

sources/standards/plugins.move
```diff
@@ -0,0 +1,59 @@
...
+    /// === Interoperability ===
+
+    public fun borrow_plugin_domain<C>(
+        collection: &Collection<C>,
+    ): &PluginDomain {
+        collection::borrow_domain(collection)
+    }
+
+    /// Requires that sender is a creator
+    public fun borrow_plugin_domain_mut<C, W: drop>(
```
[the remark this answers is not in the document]
bausano: Exposing both now

sources/utils/multisig.move
bausano: - Polished the `Multisig` API to construct it from an arbitrary `VecMap<address, u16>`.
- It's not droppable as it's going to be a shared object.
- Multisig out of creators is a convenience, saying "3 out of 5 creators must agree" is nice.
- Alternatively, I could make someone a creator which gives them lots of administrative power, but I might not want to let them to publish completely arbitrary logic but stay within bounds of what's been deployed, which can be achieved with this multisig.

bausano:
> PluginDomain seems unclear to me in terms of naming, somehting like AuthorityDomain would make immediate sense.
> 
> As part of this PR I think we should take the opportunity to revamp CreatorsDomain which has unclear and conflicting responsibilities with this PR.
> 
>     * Remove CreatorsDomain outright in favor of AuthorityDomain and moving royalty share allocation to RoyaltyDomain
> 
>     * AuthorityDomain is used to authenticate mutable operations on OriginByte domains
> 
>     * AuthorityDomain should be capable of authenticating both address and witness (big win imo)
> 
>     * All domains should return two methods `borrow_domain_mut_with_address` `borrow_domain_mut_with_witness` (so methods like `display::set_name(collection: &mut Collection<C>, name: String)` should be replaced with unprotected `display::set_name(domain: &mut DisplayDomain, name: String)`.

- Alternative naming is `ExtensionDomain` or `AddonDomain`, although I believe also `PluginDomain` expresses the pattern well enough: it's about smart contracts which can "pretend" to be the main contract that contains the OTW.
- I agree that all domains should return two versions of the method, that's a good idea we should retroactively implement.

Move.toml
```diff
@@ -14,4 +14,4 @@ git = "https://github.com/Origin-Byte/originmate.git"
 rev = "7dd2b3a6add03d140cdbc48e7c2a0e546ba24047"
 
 [addresses]
-nft_protocol = "0x1d84213ee71d5f3948435df9e5b20a5d724c88d1"
+nft_protocol = "0x0"
```
bausano, line 17: TODO

nmboavida (approved):

nmboavida (approved):

PR #192, opened 2023-01-04
Document: Collection

Suficio:
Documentation for Collection

* NFT docs polishing
* Hotfix NFT API
* Improved domain error reporting
* Test support for published packages


sources/nft/nft.move
```diff
@@ -54,10 +63,14 @@ module nft_protocol::nft {
     /// struct SUIMARINES has drop {}
     ///
     /// fun init(witness: SUIMARINES, ctx: &mut TxContext) {
-    ///     let nft = nft::new<SUIMARINES>(tx_context::sender(ctx), ctx);
+    ///     let nft = nft::new(&witness, tx_context::sender(ctx), ctx);
     /// }
     /// ```
-    public fun new<C>(owner: address, ctx: &mut TxContext): Nft<C> {
+    public fun new<C>(
+        _witness: &C,
```
bausano, line 70: I'd make witness `W` and then used the `assert_witness_same_package` because otherwise we could only create NFTs in the `init` function.

Suficio: truee 🥝

Suficio: Done

nmboavida (approved):

PR #214, opened 2023-01-16
Feature:  SupplyDomain

Suficio:


sources/launchpad/inventory.move
```diff
@@ -1,237 +1,235 @@
...
+    /// Increments the delegated supply of `Inventory`
+    ///
+    /// This endpoint must be called before a new `Nft` object is created to
+    /// ensure that global supply tracking remains consistent.
+    ///
+    /// #### Panics
+    ///
+    /// Panics if `Inventory` is unregulated.
+    public entry fun increment_supply<C>(
+        inventory: &mut Inventory<C>,
```
bausano, line 207: Is the idea that `&mut Inventory` cannot be acquired without some sort of admin permission? What are the scenarios under which one has an access to `&mut Inventory`?

sources/launchpad/inventory.move
```diff
@@ -1,237 +1,235 @@
...
-
-    /// Get specific `Inventory` market
-    public fun market<Market: key + store>(
-        inventory: &Inventory,
-        market_id: ID,
-    ): &Market {
-        assert_market<Market>(inventory, market_id);
-        object_bag::borrow<ID, Market>(&inventory.markets, market_id)
+    /// Return how many `Nft` there are to sell
+    public fun length<C>(inventory: &Inventory<C>): u64 {
```
bausano, line 220: alternatively `size`

sources/launchpad/venue.move
```diff
@@ -0,0 +1,232 @@
...
+            market_id,
+            market,
+        );
+    }
+
+    /// Deposits `Inventory` to `Venue`
+    ///
+    /// Endpoint is unprotected and relies on safely obtaining a mutable
+    /// reference to `Venue`.
+    // public entry fun merge_inventory<C>(
```
bausano, line 81: TODO

sources/launchpad/inventory.move
Suficio: `Inventory` is a private and unprotected object not meant to be shared

sources/launchpad/venue.move
Suficio: Done

sources/launchpad/inventory.move
Suficio: Done

sources/launchpad/venue.move
```diff
@@ -0,0 +1,245 @@
...
+            id: object::new(ctx),
+            live: vec_map::empty(),
+            whitelisted: vec_map::empty(),
+            markets: object_bag::new(ctx),
+            nfts_on_sale: vector::empty(),
+        }
+    }
+
+    /// Creates a `Venue` and transfers to transaction sender
+    public entry fun init_inventory(ctx: &mut TxContext) {
```
nmboavida, line 62: Should this not be called `init_venue` instead of `init_inventory`?

sources/launchpad/venue.move
```diff
@@ -0,0 +1,245 @@
...
+        collection: &mut Collection<C>,
+        inventory: Inventory<C>,
+    ) {
+        while (!inventory::is_empty(&inventory)) {
+            let nft = inventory::redeem_nft(&mut inventory);
+            deposit_nft(venue, nft);
+        };
+
+        inventory::destroy(collection, inventory);
+    }
```
nmboavida, line 104: This seems quite expensive. Can we not instead transfer the Inventory to the Venue?

Suficio: No as multiple `Venues` are transferable to one `Inventory`.
I could add an extra function to simply create a `Venue` from `Inventory` if someone wants to avoid that cost.

PR #219, opened 2023-01-20
Upgraded to Sui 0.22.0

nmboavida:


Move.toml
```diff
@@ -1,17 +1,17 @@
 [package]
 name = "NftProtocol"
-version = "0.18.0"
+version = "0.19.0"
```
Suficio, line 3: We shouldn't be bumping minor version for simple dependency bumps

Suficio (approved):

Move.toml
bausano: But it does break the dependency chain though

PR #223, opened 2023-01-23
Feature: Reorganize Venue

Suficio:
- Single market per `Venue`
- `Venue` erases market type and handles administrative tasks
- `Venue` can only be assigned to a single inventory, massively simplifies end user API at negligible extra cost to setup (In most cases its even easier to setup)
- Expanded end to end tests to test different `Marketplace` setup paradigms and actually buy the NFTs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,18 @@ Changelog](https://keepachangelog.com/en/1.0.0/), and this project adheres to
 
 ### Added
 
+- `PluginDomain` which collects type names of witness structs.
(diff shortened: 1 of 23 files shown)


nmboavida:
> - Single market per `Venue`
> - `Venue` erases market type and handles administrative tasks
> - `Venue` can only be assigned to a single inventory, massively simplifies end user API at negligible extra cost to setup (In most cases its even easier to setup)
> - Expanded end to end tests to test different Marketplace setup paradigms and actually buy the NFTs

Just documenting last update here:
 - Inventories are now defined at the market level, so anyone can implement their primitives that that use multiple inventories

nmboavida (approved):

PR #233, opened 2023-02-01
Feature: NFT permissions

Suficio:
PR brings NFT permissions inline with `DelegatedWitness` changes and loose NFT minting.

In effect this leaves our permission system in the following state:

- `DelegatedWitness`
    - May add `Collection` and NFT domains
    - May modify standard domains
- `MintCap`
    - May mint NFT
    - May delegate creation of `RegulatedMintCap` and `UnregulatedMintCap`
    - May add domains to NFT but only with original contract `Witness`
        - This is so `MintCap` may be safely shared
- `UnregulatedMintCap`
    - May mint NFT
    - May add domains to NFT only if transaction sender is also `logical_owner`
    - May be transposed to `RegulatedMintCap`
- `RegulatedMintCap`
    - May mint NFT
    - May add domains to NFT only if transaction sender is also `logical_owner`
- `logical_owner`
    - May add domains to NFT if `ModifyDomain` exists
    
Additionally, `Inventory` now safely changes the `logical_owner` of redeemed NFTs using an embedded `Allowlist`.

### DIFF
--- a/sources/nft/nft.move
+++ b/sources/nft/nft.move
@@ -89,21 +89,30 @@ module nft_protocol::nft {
 
     /// Create a new `Nft`
     ///
+    /// This operation is authorized using a `MintCap` but also requires that
+    /// it is called inside the original contract.
+    /// This restriction is imposed in order to allow `MintCap` to be shared
+    /// and provide a protected API for minting NFTs.
+    /// If you want to delegate minting of NFTs outside of the defining
+    /// contract, see [new_regulated](#new_regulated) or
+    /// [new_unregulated](#new_unregulated).
+    ///
     /// #### Usage
     ///
     /// ```
     /// struct SUIMARINES has drop {}
     ///
     /// fun init(witness: SUIMARINES, ctx: &mut TxContext) {
-    ///
     ///     let nft = nft::new(&witness, tx_context::sender(ctx), ctx);
     /// }
     /// ```
-    public fun new<C>(
+    public fun new<C, W>(
+        _witness: &W,
         _mint_cap: &MintCap<C>,
         owner: address,
         ctx: &mut TxContext,
     ): Nft<C> {
+        utils::assert_same_module_as_witness<C, W>();
         new_(owner, ctx)
     }
 
--- a/sources/nft/nft.move
+++ b/sources/nft/nft.move
@@ -14,19 +14,14 @@ module nft_protocol::nft {
     use sui::dynamic_object_field as dof;
     use sui::object::{Self, ID, UID};
     use sui::transfer;
-    use sui::tx_context::{Self, TxContext};
+    use sui::tx_context::TxContext;
 
-    use nft_protocol::err;
+    use nft_protocol::witness::Witness as DelegatedWitness;
     use nft_protocol::utils::{Self, Marker};
     use nft_protocol::transfer_allowlist::{Self, Allowlist};
     use nft_protocol::mint_cap::{
         Self, MintCap, RegulatedMintCap, UnregulatedMintCap,
     };
-
-    // TODO: Remove this after refactoring NFT permissions to allow
-    // DelegatedWitness
-    friend nft_protocol::warehouse;
-    friend nft_protocol::loose_mint_cap;
 
     /// Domain not defined
     ///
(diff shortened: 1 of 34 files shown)


bausano (approved):

sources/nft/nft.move
```diff
@@ -91,21 +89,30 @@ module nft_protocol::nft {
...
     /// struct SUIMARINES has drop {}
     ///
     /// fun init(witness: SUIMARINES, ctx: &mut TxContext) {
-    ///
     ///     let nft = nft::new(&witness, tx_context::sender(ctx), ctx);
     /// }
     /// ```
-    public fun new<C>(
+    public fun new<C, W>(
+        _witness: &W,
```
bausano, line 110: I'd have one method which accepts witness, and other which accepts mint cap. So the creator can chose whether to authorize with witness or mint cap

sources/safe/transfer_allowlist.move
```diff
@@ -99,14 +129,16 @@ module nft_protocol::transfer_allowlist {
...
         _allowlist_witness: &Admin,
-        _authority: &CollectionControlCap<C>,
+        authority: CollectionControlCap<C>,
         list: &mut Allowlist,
     ) {
         assert_admin_witness<Admin>(list);
 
         vec_set::insert(&mut list.collections, type_name::get<C>());
+
+        delete_collection_cap(authority);
```
bausano, line 141: I don't see a reason to delete this one. It can be reused to add the collection to multiple allowlists.

sources/safe/transfer_allowlist.move
```diff
@@ -173,30 +205,80 @@ module nft_protocol::transfer_allowlist {
...
+    public fun contains_collection<C>(allowlist: &Allowlist): bool {
+        vec_set::contains(&allowlist.collections, &type_name::get<C>())
+    }
 
+    /// Returns whether `Allowlist` contains authority `Auth`
+    public fun contains_authority<Auth>(allowlist: &Allowlist): bool {
         if (option::is_none(&allowlist.authorities)) {
-            return applies_to_collection
-        };
+            // If no authorities are defined this effectively means that all
```
bausano, line 216: minor: semantically this is an odd return value

sources/nft/nft.move
Suficio: In this case I can add another function that only authorizes using witness, but mintcap needs additional witness verification if we want to be able to safely share the mintcap.

sources/safe/transfer_allowlist.move
Suficio: It seems counterintuitive that if collection authorized you to add it to a particular allowlist because you have audited the contracts allowed by it that the person who you transferred the control cap to can go and add it to any other arbitrary allowlist.

sources/nft/nft.move
bausano: 1. What is the point then of using the mint cap if witness is enough?
2. Isn't it the case that the mint cap would be wrapped in an object and the access to the mint cap happens via some permission module.

Perhaps, could you show some usecases where both witness and mintcap makes sense. Or where mintcap itself is not enough to do proper authorization

sources/safe/transfer_allowlist.move
bausano: I see, we both understood the semantics of the `CollectionCap` differently then. I imagined it being something owned by the collection creators who'd go via some sort of entry function in the particular whitelist implementation to add their collection. The advantage of this is that collection creators don't have to wait for allowlist admin action.

You understood it as something collection creators transfer to the allowlist admin and they perform the insertion themselves.

Now of course it depends on the flow, both are possible with or without the deletion of the collection cap afterward.

sources/nft/nft.move
Suficio: 1. Yea this doesn't make too much sense, in this case only the function that requires MintCap should be kept as it ensures that minting permissions are correctly handled in upstream modules.

2. In cases where MintCap is shared you cannot allow people to arbitrarily mint any kind of NFT, and need to enforce NFT mints only allowed through a function in the contract.
https://github.com/Origin-Byte/nft-protocol/pull/239

sources/safe/transfer_allowlist.move
Suficio: So I'll revert the changes and lets revisit this again.

sources/nft/nft.move
Suficio: `add_domain_with_mint_cap` should also take `W` which I see it doesnt for cases where `MintCap` is shared.

Suficio: Now that I think about it, MintCap should not be shared which would allow us to make a separate endpoint for MintCap and Witness.

nmboavida (approved):

PR #243, opened 2023-02-08
Feature: ComposableUrlDomain

Suficio:

PR #251, opened 2023-02-13
Feature: Compose `Nft.url`

Suficio:
Alternative to https://github.com/Origin-Byte/nft-protocol/pull/243.

Composes `UrlDomain` and `AttributesDomain` into `Nft.url` for best display compatibility.

### DIFF
--- a/sources/standards/composability/url.move
+++ b/sources/standards/composability/url.move
@@ -1,0 +1,215 @@
+/// Module of `ComposableUrlDomain`
+///
+/// `ComposableUrlDomain` does not itself compose NFTs but serves as a display
+/// standard provider for NFTs which register `UrlDomain` and
+/// `AttribtuesDomain` and are composed within `NftBagDomain`.
+module nft_protocol::composable_url {
+    use std::ascii;
+    use std::vector;
+
+    use sui::object::ID;
+    use sui::url::Url;
+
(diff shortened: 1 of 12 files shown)


bausano (approved):

bausano:
There's still going to be the URL also on the `Nft` type. How do we instruct users to know the distinction and when to use domains vs properties?

Suficio:
> There's still going to be the URL also on the `Nft` type. How do we instruct users to know the distinction and when to use domains vs properties?

This MR has it in such a way that `url` is updated on updating either `UrlDomain` or `ComposableUrlDomain` so it gives a "most correct" URL at any time.

Eitherway after the new display standards the field will dissapear and the user may choose which URL domain to use to populate the URL field.

sources/standards/composability/url.move
```diff
@@ -0,0 +1,215 @@
...
+    ) {
+        let domain = borrow_domain_mut(nft);
+        domain.url = url;
+
+        url::set_url(witness, nft, url);
+    }
+
+    /// Registers NFT whose SVG data should be composed within the final
+    /// composed SVG
+    ///
```
nmboavida, line 84: TODO: Update comments.

I assume this comes from `svg.move`?

nmboavida (approved):

PR #252, opened 2023-02-14
Feature: Composable SVG

Suficio:


sources/standards/composability/container.move
```diff
@@ -0,0 +1,226 @@
...
+        /// Type name of the witness authorized to withdraw the NFT
+        witness: TypeName,
+        /// ID of the NFT stored as a dynamic object field
+        nft_id: ID,
+    }
+
+    /// `ContainerDomain` object
+    struct ContainerDomain has store {
+        /// `ContainerDomain` ID
+        id: UID,
```
bausano, line 50: Is this useful if the domain doesn't have `key` ability?

Suficio:
It's used to store NFTs as dynamic fields, guess I should rename it to
`nfts` but that's already a field. 😅

sources/standards/composability/container.move
Suficio: It's used to store NFTs as dynamic fields, guess I should rename it to
`nfts` but that's already a field. 😅

bausano: I see, yeah I'd rename it or add a comment

bausano (approved):

sources/standards/composability/container.move
nmboavida: Renamed it to `NftBagDomain`

PR #269, opened 2023-02-21
Endpoint to list multiple NFTs in one tx

bausano:
### Added

- Helper endpoints in orderbook which allow marketplaces to perform actions in a single tx.
  The marketplaces can now
  - `list_multiple_nfts`

### Changed

- Moved errors from `err` module into `orderbook` module where they are expressed as constants.

### DIFF
--- a/sources/trading/orderbook.move
+++ b/sources/trading/orderbook.move
@@ -52,20 +52,30 @@ module nft_protocol::orderbook {
 
     // === Errors ===
 
+    /// A protected action was called without a witness.
+    /// This action can only be called from an implementation in the collection
+    /// smart contract.
     const EACTION_NOT_PUBLIC: u64 = 0;
 
+    /// # of nfts and # of requested prices must match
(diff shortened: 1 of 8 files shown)


sources/trading/orderbook.move
```diff
@@ -51,6 +50,26 @@ module nft_protocol::orderbook {
     use sui::transfer::{transfer, share_object};
     use sui::tx_context::{Self, TxContext};
 
+    // === Errors ===
+
+    const EACTION_NOT_PUBLIC: u64 = 0;
```
Suficio, line 55: Documentation as to why this error occurred and what action should be taken to mitigate it (call an endpoint or do something different) would be very nice here.

Suficio (approved):

sources/trading/orderbook.move
Suficio: This applies to all other errors

PR #270, opened 2023-02-21
Buying NFT from launchpad and depositing it to Safe

bausano:


sources/launchpad/market/fixed_price.move
```diff
@@ -188,9 +246,11 @@ module nft_protocol::fixed_price {
             );
 
         let owner = tx_context::sender(ctx);
-        inventory::transfer(inventory, owner, ctx);
+        let nft = inventory::redeem_nft(inventory, owner, ctx);
 
         listing::pay(listing, funds, 1);
```
Suficio, line 251: Could `listing::pay` be moved right after the `funds` definition?

sources/launchpad/market/fixed_price.move
```diff
@@ -137,7 +139,33 @@ module nft_protocol::fixed_price {
...
+    }
+
+    /// Buy NFT for non-whitelisted sale.
+    /// Deposits the NFT to a safe and transfers the ownership to the buyer.
+    ///
+    /// #### Panics
+    ///
+    /// Panics if `Venue` does not exist, is not live, or is whitelisted or
+    /// wallet does not have the necessary funds.
+    public entry fun create_safe_and_buy_nft<C, FT>(
```
Suficio, line 153: I think we could also add an endpoint `buy_nft_into_safe` (or something like that) that takes a reference to a safe that already exists.

Suficio (changes requested):
Please also replicate these endpoints for the `LimitedFixedPriceMarket`

sources/launchpad/market/fixed_price.move
bausano: This sort of leads into a question: how can a collection owner say they want the NFTs to be deposited into safe _only_. Because that is the royalty guarantee. Would they have to whitelist and build their own endpoint?

bausano: Added

sources/launchpad/market/fixed_price.move
bausano: Cannot due to mut ref on invetory

sources/launchpad/market/fixed_price.move
Suficio: Good point, could you create an issue so we make sure to go back to this?

Suficio (approved):

PR #290, opened 2023-03-08
Feature: Batch minting

nmboavida:


examples/suimarines.move
```diff
@@ -20,6 +21,11 @@ module nft_protocol::suimarines {
     use nft_protocol::collection::{Self, Collection};
     use nft_protocol::transfer_allowlist_domain;
 
+    const EWrongDescriptionLength: u64 = 1;
```
Suficio, line 24: Lets stick to the style that all other errors use

Suficio (approved):

examples/suimarines.move
nmboavida: Done

PR #293, opened 2023-03-09
Feature: Sui Safe

nmboavida:


sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+
+        cap
+    }
+
+    public fun auth_transfer<I: key + store, IW: drop, E: drop>(
+        nft_id: ID,
+        owner_cap: &OwnerCap,
+        safe: &mut SuiSafe<I>,
+        request: Request<E>,
+        _inner_witness: IW,
```
bausano, line 115: This can be just a type, we don't need to have the witness as an input until the actual transfer imho

bausano: never mind, this is the inner witness of course

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+    ) {
+        utils::assert_same_module_as_witness<I, IW>();
+
+        // TODO: Assert that contract_witness and object ID correspond? How can we even do that?
```
bausano, line 119: `& UID`

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+
+    struct TransferAuth has store, copy, drop {
+        entity: TypeName,
+        object_id: ID,
+    }
+
+    // TODO: Is it safe to have copy here?
+    // TODO: Function to create and burn request
+    struct Request<phantom E> has copy, drop {
+        object_id: ID
```
bausano, line 66: `Option<ID>`

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+        // TODO: It only needs VecMap<ID, TypeName>, at the same time it allows for
+        // unified type with field exclusive_auth
+        auths: VecMap<ID, TransferAuth>,
+        exclusive_auth: Option<TransferAuth>,
+        object_type: TypeName,
+    }
+
+    struct TransferAuth has store, copy, drop {
+        entity: TypeName,
+        object_id: ID,
```
bausano, line 60: `Option<ID>`

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+        utils::assert_same_module_as_witness<I, IW>();
+
+        let nft = get_nft_for_transfer_<I, E, T>(nft_id, request, safe);
+
+        nft
+    }
+
+    // TODO: To make it entry function we can't use Request
+    public fun delist_nft<I: key + store, IW: drop, E: drop>(
+        request: Request<E>,
```
bausano, line 228: Why do we need request here. There should be two options, either this happens with `&OwnerCap` and is entry, or is `public fun` and then it receives 

```
entity_witness: E,
entity_id: Option<&UID>
```

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+        nft_id: ID,
+        owner_cap: &OwnerCap,
+        safe: &mut SuiSafe<I>,
+        _inner_witness: IW,
+    ) {
+        utils::assert_same_module_as_witness<I, IW>();
+        assert_owner_cap(owner_cap, safe);
+        assert_has_nft(&nft_id, safe);
+
+        let ref = vec_map::get_mut(&mut safe.refs, &nft_id);
```
bausano, line 238: Cannot happen to exclusively listed transfer

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+    // // === Getters ===
+
+    public fun check_auth<E: drop>(request: &Request<E>, auth: &TransferAuth): bool {
+        let check_1 = request.object_id == auth.object_id;
+        let check_2 = type_name::get<E>() == auth.entity;
+
+        check_1 || check_2
+    }
+
+    // TODO: Should this not be protected?
```
bausano, line 298: I vote to only protect `&mut T`

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+    // TODO: We need to be consistent, do we use T or C?
```
bausano, line 299: Yeah I never know, perhaps we can use `NFT` as the generic?

bausano: I don't see a reason why to use _only_ single letter generic names when we don't use single letter variable names or class names or any other naming unless in special circumstances like indexing.

sources/safe/ob_safe.move
```diff
@@ -0,0 +1,211 @@
...
+        share_object(safe);
+        cap
+    }
+
+    public fun auth_transfer<Auth: drop, E: drop>(
+        nft_id: ID,
+        owner_cap: &OwnerCap,
+        safe: &mut SuiSafe<OriginByte>,
+        request: Request<E>,
+        // TODO: Problem is that orderbook can't get authority witness Auth.. perhaps programmable transactions can help here
```
bausano, line 93: As discussed on a call, this should be the orderbook witness

sources/safe/ob_safe.move
```diff
@@ -0,0 +1,211 @@
...
+        _authority: Auth,
+        _allowlist: &Allowlist,
+    ) {
+        // TODO: We need to perform some assertions with the allowlist
```
bausano, line 97: Probably not at this stage yet, we can do that on transfer

sources/safe/sui_safe.move
nmboavida: Changed

sources/safe/sui_safe.move
```diff
@@ -0,0 +1,342 @@
...
+            DepositEvent {
+                safe: object::id(safe),
+                nft: nft_id,
+            }
+        );
+    }
+
+    /// Use a transfer auth to get an NFT out of the `Safe`.
+    public fun transfer_nft_to_recipient<I: key + store, IW: drop, E: drop, T: key + store>(
+        request: Request<E>,
```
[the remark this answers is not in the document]
nmboavida: Agreed, however it seems we can't have `Option<&UID>` as a param, we get `Expected a single non-reference type, but found: '&(sui=0x2)::object::UID'`

sources/safe/sui_safe.move
nmboavida: That's right, however I am afraid if we have an entry function it can lead to accounting mismatched between the Safe and the respecitve trading contract, wdyt?

sources/safe/sui_safe.move
nmboavida: Done

sources/safe/sui_safe.move
nmboavida: Done. Removed `Request<E>` in favour of always having `&UID` in the function signatures.

sources/safe/sui_safe.move
nmboavida: Same

sources/safe/sui_safe.move
nmboavida: Changed it to `NFT`

PR #307, opened 2023-03-23
Feature: Quorum

nmboavida:


sources/collection/quorum.move
```diff
@@ -0,0 +1,649 @@
...
+
+    use nft_protocol::mint_cap::MintCap;
+
+    const ENOT_AN_ADMIN: u64 = 1;
+    const ENOT_A_MEMBER: u64 = 2;
+    const ENOT_AN_ADMIN_NOR_MEMBER: u64 = 3;
+    const EMIN_ADMIN_COUNT_IS_ONE: u64 = 4;
+    const EADDRESS_IS_NOT_ADMIN: u64 = 5;
+    const EQUORUM_EXTENSION_MISMATCH: u64 = 6;
+    const EINVALID_DELEGATE: u64 = 6;
```
bausano, line 79: 7

sources/collection/quorum.move
```diff
@@ -0,0 +1,649 @@
...
+    // === Events ===
+
+    struct CreateQuorumEvent has copy, drop {
+        quorum_id: ID,
+        type_name: TypeName,
+    }
+
+    // === Init Functions ===
+
+    public fun create<F>(
```
bausano, line 137: Why `F`

sources/collection/quorum.move
```diff
@@ -0,0 +1,649 @@
...
+    }
+
+    public fun create_for_extension<F>(
+        _witness: &F,
+        admins: VecSet<address>,
+        members: VecSet<address>,
+        delegates: VecSet<ID>,
+        ctx: &mut TxContext,
+    ): (Quorum<F>, ExtensionToken<F>) {
+        let id = object::new(ctx);
```
bausano, line 163: Can we call `create` here and you the output to then construct the token and return a tuple to reduce LoC?

sources/collection/quorum.move
```diff
@@ -0,0 +1,649 @@
...
+        let quorum = create(witness, admins, members, delegates, ctx);
+        transfer::share_object(quorum);
+    }
+
+    public fun singleton<F>(
+        _witness: &F,
+        admin: address,
+        ctx: &mut TxContext,
+    ): Quorum<F> {
+        let id = object::new(ctx);
```
bausano, line 194: Can we call the `create` function with the empty vecs as args to reduce LoC and have event emission in once place?

sources/collection/quorum.move
```diff
@@ -0,0 +1,649 @@
...
+            quorum.admin_count = quorum.admin_count + 1;
+        };
+    }
+
+    public entry fun vote_remove_admin<F>(
+        quorum: &mut Quorum<F>,
+        old_admin: address,
+        ctx: &mut TxContext,
+    ) {
+        assert!(quorum.admin_count > 1, EMIN_ADMIN_COUNT_IS_ONE);
```
bausano, line 231: `== 0`

sources/collection/quorum.move
nmboavida: Choose `F` from Family

sources/collection/quorum.move
nmboavida: Done

sources/collection/quorum.move
nmboavida: Done

sources/collection/quorum.move
nmboavida: Resolved

PR #308, opened 2023-03-23
Feature: Access Policies

nmboavida:


sources/collection/access_policy.move
```diff
@@ -0,0 +1,156 @@
...
+        utils::assert_same_module_as_witness<T, W>();
+
+        version_counter.version = version_counter.version + 1;
+
+        let id = object::new(ctx);
+
+        event::emit(NewPolicyEvent {
+            policy_id: object::uid_to_inner(&id),
+            type_name: type_name::get<T>(),
+            version: version_counter.version,
```
bausano, line 75: There'll never be an access policy with version 0, first policy starts with 1. Perhaps worth commenting somewhere in case someone is debugging and searching for the first policy.

sources/collection/access_policy.move
```diff
@@ -0,0 +1,156 @@
...
+    // get_mutable_access_by_sender
+    // get_mutable_access_by_uid
+    // get_mutable_access_by_token
+    // get_mutable_access_by_ott
+
+    struct AccessPolicy<phantom T: key + store> has key, store {
+        id: UID,
+        version: u64,
+        rules: Rules,
+        owner_sign_off: bool,
```
bausano, line 23: This will be modified to `true` in one of the functions in the TODO block?

PR #309, opened 2023-03-23
Feature: NFT Display Factory

nmboavida:
- Assert Package instead of Module
- Removed CollectionID domain
- Cleanups & Nit
- Type Extensions
- Loose NFT example with Display Factory
- Revert MintCap to use DelegatedWitness


examples/display/deadbytes.move
```diff
@@ -0,0 +1,97 @@
...
+        );
+
+        let publisher = package::claim<DEADBYTES>(witness, ctx);
+
+        transfer::transfer(mint_cap, tx_context::sender(ctx));
+        transfer::transfer(publisher, tx_context::sender(ctx));
+        transfer::share_object(collection);
+    }
+
+    // public fun mint_metadata<T: key> (
```
bausano, line 67: TODO

PR #313, opened 2023-03-29
Kiosk and transfer policy

bausano:


sources/safe/transfer_policy.move
```diff
@@ -0,0 +1,324 @@
...
+            originator,
+            ft_paid_amounts,
+            receipts: vec_set::empty(),
+        }
+    }
+
+    /// The transfer request can be converted to the sui lib version if the
+    /// payment was done in SUI and there's no other currency used.
+    ///
+    /// Note that after this, the royalty enforcement is modelled after the
```
Suficio, line 128: Not sure what this comment means, perhaps elaborate a bit more on the difference?

sources/safe/transfer_policy.move
```diff
@@ -0,0 +1,324 @@
...
+    const ENotOwner: u64 = 4;
+    /// Trying to `withdraw` more than there is.
+    const ENotEnough: u64 = 5;
+    /// Conversion of our transfer request to the one exposed by the sui library
+    /// is only permitted for SUI token.
+    const EOnlyTransferRequestOfSuiToken: u64 = 6;
+
+    /// A "Hot Potato" forcing the buyer to get a transfer permission
+    /// from the item type (`T`) owner on purchase attempt.
+    struct TransferRequest<phantom T> {
```
Suficio, line 46: From my understanding this is different than the Sui version, could you elaborate in docs what the difference is?

sources/safe/transfer_policy.move
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    /// Allow a `TransferRequest` for the type `T`. The call is protected
+    /// by the type constraint, as only the publisher of the `T` can get
+    /// `TransferPolicy<T>`.
+    ///
+    /// Note: unless there's a policy for `T` to allow transfers,
+    /// Kiosk trades will not be possible.
+    public fun confirm_request<T>(
+        self: &TransferPolicy<T>, request: TransferRequest<T>
```
Suficio, line 202:
```suggestion
        self: &TransferPolicy<T>,
        request: TransferRequest<T>,
```

Suficio (approved):

sources/safe/transfer_policy.move
bausano: Added module docs:

```
/// The rolling hot potato pattern was designed by us in conjunction with
/// Mysten team.
///
/// To lower the barrier to entry, we mimic those APIs where relevant.
/// See the `sui::transfer_policy` module in the https://github.com/MystenLabs/sui
/// We interoperate with the sui ecosystem by allowing our `TransferRequest` to
/// be converted into the sui version.
/// This is only possible for the cases where the payment is done in SUI token.
///
/// Our transfer request offers generics over fungible token.
/// We are no longer limited to SUI token.
/// Royalty policies can decide whether they charge a fee in other tokens.
///
/// Our transfer request associates paid balance with the request.
/// This enables us to do permissionless transfers of NFTs.
/// That's because we store the beneficiary address (e.g. NFT seller) with the
/// paid balance.
/// Then when confirming a transfer, we transfer this balance to the seller.
/// With special capability, the policies which act as a middleware can access
/// the balance and charge royalty from it.
/// Therefore, a 3rd party to a trade can send a tx to finish it.
/// This is helpful for reducing # of txs users have to send for trading
/// logic which requires multiple steps.
/// With our protocol, automation can be set up by marketplaces.
```

sources/safe/transfer_policy.move
bausano:
```
    /// Note that after this, the royalty enforcement is modelled after the
    /// sui ecosystem settings.
    /// This means that the rules for closing the `SuiTransferRequest` must
    /// be met according to `sui::transfer_policy::confirm_request`.
```

sources/launchpad/market/fixed_price.move
```diff
@@ -151,34 +151,16 @@ module nft_protocol::fixed_price {
...
-    }
-
-    /// Buy NFT for non-whitelisted sale.
-    /// Deposits the NFT to a safe and transfers the ownership to the buyer.
-    ///
-    /// #### Panics
-    ///
-    /// Panics if `Venue` does not exist, is not live, or is whitelisted or
-    /// wallet does not have the necessary funds.
-    public entry fun create_safe_and_buy_nft<C, FT>(
```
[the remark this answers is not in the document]
bausano: With programmable txs this can be done client side.

sources/safe/transfer_policy.move
```diff
@@ -0,0 +1,324 @@
...
+        // since we unwrap above to get the SUI amount, we know that there're no
+        // other currencies
+        assert!(vec_map::size(&ft_paid_amounts) == 1, EOnlyTransferRequestOfSuiToken);
+
+        sui_transfer_policy::new_request(
+            nft, paid, object::id_from_address(originator)
+        )
+    }
+
+    /// Register a type in the Kiosk system and receive an `TransferPolicyCap`
```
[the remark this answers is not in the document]
bausano: Removed this type

sources/safe/transfer_policy.move
```diff
@@ -0,0 +1,324 @@
...
+
+        (
+            TransferPolicy { id, rules: vec_set::empty() },
+            TransferPolicyCap { id: object::new(ctx), policy_id },
+        )
+    }
+
+    /// Withdraw some amount of profits from the `TransferPolicy`.
+    /// If amount is not specified, all profits are withdrawn.
+    public fun withdraw<T, FT>(
```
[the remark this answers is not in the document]
bausano: I removed the withdraw method in favor of the implementation of `sui::transfer_policy::TransferPolicy`. Their version doesn't have a generic of FT. Unfortunately, they protect the access to `&mut UID` with a `TransferPolicyCap`. Alas, this means we cannot access dynamic fields.
The royalty collection for FT is going to happen outside of this module _somehow_, either via our old `&mut Collection` version or something new.
I removed this responsibility from this file.

sources/safe/transfer_policy.move
bausano: (Where is `move-fmt`? Thinking about micro formatting is so 2000 and 15.)

While I generally prefer this too, I adopted the formatting style of `sui` dependency. When formatting, I am aiming for the smallest diff to that dependency. When a formatting tool comes, I expect it will follow those rules. Although, I personally don't like their style visually pleasing.

sources/safe/transfer_policy.move
```diff
@@ -0,0 +1,324 @@
...
+    struct TransferRequestBuilder<phantom T> {
+        nft: ID,
+        originator: address,
+        ft_paid_amounts: VecMap<TypeName, u64>,
+    }
+
+    /// A unique capability that allows owner of the `T` to authorize
+    /// transfers.
+    /// Can only be created with the `Publisher` object.
+    struct TransferPolicy<phantom T> has key, store {
```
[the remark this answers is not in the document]
bausano: We discussed this briefly offhub (look at me coming up with new words.)
More docs will follow shortly during the week.

nmboavida (approved):

PR #316, opened 2023-04-03
Feature: Domain access

Suficio:


nmboavida:
I am wondering if we should use DfKeys instead of Markers when fields are in domains. It's not clear to me yet, but we need to consider if the creator has legitimate reasons for sharing `&mut UID`.  Let's consider the difference between the two patterns (Using Domain Permissions vs. Consumable Witness) and how to juxtapose to each other.

PR #327, opened 2023-04-04
`FrozenPublisher` as a solution for `Display`

bausano:


sources/utils/frozen_publisher.move
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    public fun mod(self: &FrozenPublisher): &ascii::String {
+        package::published_module(&self.inner)
+    }
+
+    public fun borrow_publisher<T>(
+        _witness: DelegatedWitness<T>, self: &FrozenPublisher,
+    ): &Publisher {
+        assert!(package::from_module<T>(&self.inner), 0);
```
bausano, line 58: TODO: errors

examples/suimarines.move
```diff
@@ -51,7 +51,9 @@ module nft_protocol::suimarines {
             &transfer_policy_cap,
         );
 
-        royalty_strategy_bps::create_domain_and_add_strategy<SUIMARINES>(
+        let delegated_witness = witness::from_witness<SUIMARINES, Witness>(Witness {});
```
Suficio, line 54:
```suggestion
        let delegated_witness = witness::from_witness(Witness {});
```

sources/standards/royalties/royalty_strategy.move
```diff
@@ -1,22 +1,30 @@
-module nft_protocol::royalty_strategy_bps {
-    use std::fixed_point32;
-    use std::option::{Self, Option};
-
+module nft_protocol::bps_royalty_strategy {
```
Suficio, line 1: Would prefer having the naming order in the old scheme, it orders the name in terms of hierarchal responsibility:
`nft_protocol/royalty_strategy/bps => nft_protocol::royalty_strategy_bps`
`nft_protocol/tests/mint_cap => nft_protocol::tests_mint_cap`

Nit pick but remainder of code separates import groups by package, would be nice to retain that pattern.

sources/utils/frozen_publisher.move
```diff
@@ -0,0 +1,84 @@
...
+        &self.inner
+    }
+
+    /// FrozenPublisher has Publisher from OTW of PARENT.
+    ///
+    /// PARENT: a::b::Foo<c::d::Bar<..>>
+    /// INNER: c::d::Bar<..>
+    ///
+    /// Asserts that inner type of Foo equals to the Bar.
+    public fun new_display_for_inner_generic<PARENT: key, INNER>(
```
Suficio, line 68:
```suggestion
    public fun new_display_for_inner_generic<Parent: key, Inner>(
```

sources/utils/frozen_publisher.move
```diff
@@ -0,0 +1,84 @@
...
+        _witness: DelegatedWitness<INNER>, self: &FrozenPublisher, ctx: &mut TxContext,
+    ): Display<PARENT> {
+        let (_, _, parent) = utils::get_package_module_type<PARENT>(); // Foo<c::d::Bar<..>>
+        let open_chicken = string::index_of(&parent, &utf8(b"<"));
+
+        let inner = sub_string(&parent, open_chicken + 1, string::length(&parent) - 1); // c::d::Bar<..>
+
+        let (pkg, mod, typ) = utils::get_package_module_type_raw(inner);
+        let (expected_pkg, expected_mod, expected_typ) = utils::get_package_module_type<PARENT>();
```
Suficio, line 77: Should this not be `INNER`?

```suggestion
        let (expected_pkg, expected_mod, expected_typ) = utils::get_package_module_type<INNER>();
```

sources/utils/frozen_publisher.move
```diff
@@ -0,0 +1,84 @@
...
+    /// Asserts that inner type of Foo equals to the Bar.
+    public fun new_display_for_inner_generic<PARENT: key, INNER>(
+        _witness: DelegatedWitness<INNER>, self: &FrozenPublisher, ctx: &mut TxContext,
+    ): Display<PARENT> {
+        let (_, _, parent) = utils::get_package_module_type<PARENT>(); // Foo<c::d::Bar<..>>
+        let open_chicken = string::index_of(&parent, &utf8(b"<"));
+
+        let inner = sub_string(&parent, open_chicken + 1, string::length(&parent) - 1); // c::d::Bar<..>
+
+        let (pkg, mod, typ) = utils::get_package_module_type_raw(inner);
```
Suficio, line 76: I would expect that inner could potentially be `c::d::Bar<..>, .., ..`

I think it would be safer to assert `inner` length to length of `type_name::get<INNER>()` and compare the strings. Would also save quite a few string operations.

sources/utils/frozen_publisher.move
```diff
@@ -0,0 +1,84 @@
...
+    use std::string::{Self, utf8, sub_string};
+    use sui::display::{Self, Display};
+    use sui::object::{Self, UID};
+    use sui::package::{Self, Publisher};
+    use sui::transfer;
+    use sui::tx_context::TxContext;
+
+    struct FROZEN_PUBLISHER has drop {}
+
+    struct FrozenPublisher has key {
```
Suficio, line 14: I'm not clear why we need `FrozenPublisher` as a type instead of a function that just takes `Publisher`?

nmboavida:
Copying Michael's comments from Slack here:

This creates a helper which wraps `Publisher`. 

Useful for packages which have generics but want to export Display.

For example, `BpsRoyaltyStrategy<T>` cannot set a `Display<BpsRoyaltyStrategy<INNER>>` at once. For each `INNER` it must be called again.

This allows to get a `Display<BpsRoyaltyStrategy<INNER>>`  based on `DelegatedWitness<INNER>`.


**What Frozen Publisher does:**

Wraps publisher and shares it immutably.

The main use case right now is getting a `Display` object for inner types of our protocol. Sui might eventually create a solution for setting Display of generics. Perhaps this won't be needed.

Have a look at the `sui::display module`. To create `Display`, you need access to `Publisher`.
`Display<Orderbook<T, FT>>` cannot be created, you ought to know specific type.

nmboavida (approved):

PR #331, opened 2023-04-05
Hotfix: MintCap

Suficio:


sources/collection/mint_event.move
```diff
@@ -77,7 +77,6 @@ module nft_protocol::mint_event {
     ) {
         // Assert that there is a supply in mint cap
         mint_cap::assert_regulated(mint_cap);
-        assert!(mint_cap::is_frozen(mint_cap), 0);
```
nmboavida, line 80: Why is this being deleted?

`assert!(mint_cap::is_frozen(mint_cap), 0);`

This should only be called if the mint cap is frozen, that is, the supply inside of it is frozen.

sources/collection/mint_cap.move
```diff
@@ -141,16 +116,17 @@ module nft_protocol::mint_cap {
         supply::get_current(option::borrow(&mint_cap.supply))
     }
 
-    public fun is_frozen<T>(mint_cap: &MintCap<T>): bool {
-        let supply = get_supply(mint_cap);
-        supply::is_frozen(supply)
-    }
-
```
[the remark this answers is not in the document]
Suficio: Because its always set to frozen.

```
// The supply is always set to frozen for safety
supply: option::some(supply::new(supply, true)),
```

sources/collection/mint_event.move
Suficio: Same as above, because its always frozen.

nmboavida (approved):

PR #335, opened 2023-04-05
Feature: Composability fixes

Suficio:
* Bring `Composition` domain interface inline with others after rewrite
* Remove type restriction from `NFtBag`, this was needlessly strict and there is no situation where you don't know the type of the NFT you're withdrawing like Warehouse.

### DIFF
--- a/examples/tribal_realms.move
+++ b/examples/tribal_realms.move
@@ -73,10 +73,10 @@ module nft_protocol::tribal_realms {
         // === Avatar composability ===
 
         let avatar_blueprint = c_nft::new_composition<Avatar>();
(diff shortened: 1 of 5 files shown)


nmboavida:
We still need to have an interface for accessing the functionality via the `&mut UID `of the parent objects, because otherwise we force interface reimplementation on the part of the Creator. This is because if the whole interface expects `NftBag`, and given that NftBag lives inside parent NFTs, then we need a way to access this child object, the only way to do that is to access it via the scope of a function that creator implements that itself takes the parent as a parameter.

I will merge this MR for now and readd the UID interface for all the domains in a subsequent MR.

nmboavida (approved):

PR #336, opened 2023-04-06
Helpers: Access policies

nmboavida:


sources/permissions/access_policy.move
```diff
@@ -1,4 +1,8 @@
 module nft_protocol::access_policy {
+    // TODO: Consider accessors to Kiosk
```
bausano, line 2: What are the benefits? Should we create an issue for this?

nmboavida: The TODO msg is a bit misleading I guess. What would be good is to have the NFT owner the ability to issue One Time Mutability Tokens. The idea is to have users issue this One Time tokens to game backends at the beginning of a gaming session.

PR #350, opened 2023-04-11
`Testract` contract with various tests and examples

bausano:
This contract helps you generate some events and txs.
It can be used to test client logic.

It also contains examples on how to use certain aspects of the protocol,
especially trading.
See `generate_bidding_events`.

It's written such that a single sender calls the entry functions in the same
order that they are declared.
Read the created objects from those txs as you'll need some as inputs to the
next transactions.
We start the module with a `it_works` test which shows how those txs should
be called in order.


example_collections/examples/sources/testract.move
```diff
@@ -0,0 +1,597 @@
...
+
+    // This is the creator, buyer and seller - all in one.
+    #[test_only]
+    const USER: address = @0xA1C04;
+
+    // Calls all logic in this test package.
+    //
+    // See the function docs for more information on how they fit in.
+    #[test]
+    fun it_works() {
```
Suficio, line 499: Please use a more module-specific name like perhaps `test_example_testract` so that it can be isolated when filtering tests.

Suficio (approved):

example_collections/examples/sources/testract.move
bausano: Renamed

PR #369, opened 2023-04-18
Feature: Generic Request/Policy

nmboavida:


sources/permissions/mut_lock.move
```diff
@@ -24,6 +25,18 @@ module nft_protocol::mut_lock {
         field: Option<TypeName>,
     }
 
+    struct SessionToken<phantom T> has key {
```
Suficio, line 28: What is the goal of `SessionToken`?

Thus far it seems like it is just a more complicated version of obtaining a `MutLock`.

It should not be included in this PR util a use case and example is established.

sources/permissions/access_policy.move
```diff
@@ -221,31 +204,39 @@ module nft_protocol::access_policy {
...
         );
 
         assert!(
-            vec_set::contains(&access_policy.parent_access, &tx_context::sender(ctx)),
+            vec_set::contains(&access_policy.parent_access, &entity),
             EPARENT_ACCESS_DENIED
         );
     }
 
+    public fun assert_no_owner_signoff<T: key + store>(
```
Suficio, line 221: Where is this going to be used?

If unused it should not be in this PR.

sources/kiosk/request.move
```diff
@@ -0,0 +1,287 @@
...
+            receipts: vec_set::empty(),
+        }
+    }
+
+    /// Aborts unless called exactly once.
+    public fun set_paid<FT>(
+        self: &mut RequestBody, paid: Balance<FT>, beneficiary: address,
+    ) {
+        self.beneficiary = beneficiary;
+        df::add(&mut self.metadata, BalanceDfKey {}, paid);
```
Suficio, line 137: This should be a plain `Option`, otherwise you're introducing the overhead of dynamic fields for something that really doesn't need them.

sources/kiosk/request.move
```diff
@@ -0,0 +1,287 @@
...
+        /// Originator of the request. For entities which authorize with `ID`,
+        /// we convert the `ID` to address.
+        originator: address,
+        /// Who's to receive the payment which we wrap as dyn field in metadata.
+        beneficiary: address,
+        /// Collected Receipts.
+        ///
+        /// Used to verify that all of the rules were followed and
+        /// `TransferRequest` can be confirmed.
+        receipts: VecSet<TypeName>,
```
Suficio, line 68: This should be made a vector as discussed.

sources/kiosk/request.move
```diff
@@ -0,0 +1,287 @@
...
+    public fun new(
+        object: ID, originator: address, ctx: &mut TxContext,
+    ): RequestBody {
+        RequestBody {
+            metadata: object::new(ctx),
+            object,
+            originator,
+            // is overwritten in `set_paid` if any balance is associated with
+            beneficiary: @0x0,
+            receipts: vec_set::empty(),
```
Suficio, line 128: Why not just immediately require that all receipts are inserted here and then progressively removed?

I think this would allow us to remove `Policy`.

Suficio: It would also make it dead easy for contracts to find out what authorization they need to obtain next as it will simply be the next one in the vector, rather than having to figure it out from the structure that initially issued the request.

PR #375, opened 2023-04-19
Kiosk Policies

nmboavida:


sources/kiosk/policies/borrow.move
```diff
@@ -0,0 +1,81 @@
...
+
+    // === Structs ===
+
+    struct Witness has drop {}
+    struct BORROW_REQUEST has drop {}
+
+    struct BorrowRequest<T> {
+        nft: T,
+        sender: address,
+        field: Option<TypeName>,
```
bausano, line 23: could you add a comment for what does this bit work do please

sources/kiosk/policies/borrow.move
```diff
@@ -0,0 +1,81 @@
...
+
+    /// Adds a `Receipt` to the `Request`, unblocking the request and
+    /// confirming that the policy requirements are satisfied.
+    public fun add_receipt<T, Rule>(self: &mut BorrowRequest<T>, rule: &Rule) {
+        request::add_receipt(&mut self.inner, rule);
+    }
+
+    public fun inner_mut<T>(
+        self: &mut BorrowRequest<T>
+    ): &mut RequestBody<WithNft<T, BORROW_REQUEST>> { &mut self.inner }
```
bausano, line 57: Nice, I think it's gonna work great with allowlist, it can now be reused for `OB_TRANSFER_REQUEST` and `BORROW_REQUEST` or `WITHDRAW_REQUEST`.

bausano (approved):

Suficio (changes requested):
Not sure that I see the benefit of separating `TransferRequest` and `WithdrawalRequest`.

Intuitively the only difference between the two is that `TransferRequest` is a `WithdrawalRequest` that requires a final receipt of being deposited in another `Kiosk`.

Suficio (approved):

sources/kiosk/policies/borrow.move
nmboavida: We will reorganize things a little bit later today and add comments

PR #379, opened 2023-04-19
Make P2P Transfers Great Again

nmboavida:


bausano:
We've already got an implementation for this

```
    /// If both kiosks are owned by the same user, then we allow free transfer.
    public fun transfer_between_owned<T: key + store>(
        source: &mut Kiosk,
        target: &mut Kiosk,
        nft_id: ID,
        ctx: &mut TxContext,
    ) {
        assert_permission(source, ctx);
        // could result in a royalty free trading by everyone wrapping over our
        // kiosk
        assert!(kiosk::owner(source) != PermissionlessAddr, ENotAuthorized);
        // both kiosks are owned by the same user
        assert!(kiosk::owner(source) == kiosk::owner(target), ENotOwner);

        let refs = df::borrow_mut(ext(source), NftRefsDfKey {});
        let ref = table::remove(refs, nft_id);
        assert_ref_not_exclusively_listed(&ref);

        let cap = pop_cap(source);
        let nft = kiosk::take<T>(source, &cap, nft_id);
        set_cap(source, cap);

        deposit(target, nft, ctx);
    }
```

bausano:
If we can use the `RequestBody` pattern for this, such that it can be disabled by creators, this is a great solution indeed!

PR #387, opened 2023-04-24
Feature: Accessible `Allowlist`

Suficio:


contracts/allowlist/sources/allowlist.move
```diff
@@ -0,0 +1,24 @@
...
+        let (al, al_cap) = transfer_allowlist::new(ctx);
+
+        // orderbooks can perform trades with our allowlist
+        transfer_allowlist::insert_authority<orderbook::Witness>(&al_cap, &mut al);
+        // bidding contract can perform trades too
+        transfer_allowlist::insert_authority<bidding::Witness>(&al_cap, &mut al);
+
+        // Delete `AllowlistOwnerCap` to guarantee that each release of
+        // `OriginByte` always has a fixed set of trading contracts
+        transfer_allowlist::delete_owner_cap(al_cap);
```
nmboavida, line 21: We should retain the allowlist cap, it will be important to add other parties in the ecosystem as they appear, without us having to update our protocol

PR #391, opened 2023-04-24
Trade resolution can happen with NFTs which have been locked

nmboavida:


nmboavida:
Merged in another branch

PR #417, opened 2023-04-30
Fix: return correct mut State for "rule_state_mut"

zengxinhai:
The "rule_state_mut" is meant to return mut object, but it's immutable currently.


nmboavida (approved):

nmboavida:
Great catch, thanks!

PR #419, opened 2023-04-30
Dynamic Vectors

nmboavida:


contracts/utils/sources/dynamic_vec.move
```diff
@@ -0,0 +1,545 @@
...
+    use originmate::pseudorandom;
+
+    #[test_only]
+    fun destroy_test<Element: store + drop>(v: DynVec<Element>) {
+        let DynVec<Element> { vec_0: _, vecs, current_chunk: _, tip_length: _, total_length: _, limit: _ } = v;
+
+        object::delete(vecs);
+    }
+
+    #[test]
```
Suficio, line 316: Tests only test that you can push and pop, can you migrate the tests from `launchpad::warehouse::Warehouse` to demonstrate that you are also popping the correct element?

contracts/utils/sources/dynamic_vec.move
```diff
@@ -0,0 +1,545 @@
...
+
+    const ECAPACITY_DECREASE_EXCEEDS_LENGTH: u64 = 3;
+
+    /// Create an empty dynamic vector.
+    public fun empty<Element: store>(limit: u64, ctx: &mut TxContext): DynVec<Element> {
+        DynVec {
+            vec_0: vector::empty(),
+            vecs: object::new(ctx),
+            current_chunk: 0,
+            tip_length: 0,
```
Suficio, line 59: Don't understand why you need to separately track and store the `current_chunk` and `tip_length` variables.

You can calculate them using the chunk index as long as you know the length
`(current_chunk, tip_length - 1) = chunk_index(v, v.total_length - 1)`

Suficio (changes requested):

contracts/utils/sources/dynamic_vec.move
nmboavida: Done

contracts/utils/sources/dynamic_vec.move
nmboavida: It's more efficient to have them stored instead of having to recompute each time, and they require not a lot of storage. It also helped me decrease the cognitive load

PR #435, opened 2023-05-02
Launchpad internal functions `friend`-protected

nmboavida:


contracts/launchpad/sources/launchpad/market/market_whitelist.move
```diff
@@ -66,7 +67,7 @@ module ob_launchpad::market_whitelist {
     }
 
     /// Burns a `Certificate`
-    public entry fun burn(
+    public fun burn(
```
Suficio, line 70: Why remove entry?

nmboavida: Users can still call this on the client side without entry, so we're good

PR #475, opened 2023-06-02
Feature:  Kiosk improvements

Suficio:
- Fully documented most functions, some function locations are re-arranged
- Improved error reporting throughout
- Quality of life improvements to working with locked NFTs

### DIFF
--- a/contracts/kiosk/sources/kiosk/ob_kiosk.move
+++ b/contracts/kiosk/sources/kiosk/ob_kiosk.move
@@ -769,7 +769,7 @@ module ob_kiosk::ob_kiosk {
         ctx: &mut TxContext,
     ) {
         assert!(kiosk::has_access(self, &kiosk_cap), ENotOwner);
-        assert!(!is_ob(self), EKioskOriginByteVersion);
+        assert!(!is_ob_kiosk(self), EKioskOriginByteVersion);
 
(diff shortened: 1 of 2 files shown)


nmboavida (approved):
When rebasing please don't forget to check that all relevant endpoints are calling assert_and_upgrade instead of just assert_version as I believe the new functions will not rebase this loc

contracts/kiosk/sources/kiosk/ob_kiosk.move
```diff
@@ -563,50 +697,58 @@ module ob_kiosk::ob_kiosk {
...
         self: &mut Kiosk,
         owner_token: OwnerToken,
         ctx: &mut TxContext,
     ) {
         assert_version_and_upgrade(ext(self));
-        assert!(owner_token.kiosk == object::id(self), EIncorrectOwnerToken);
-        assert_owner_address(self, sender(ctx));
 
-        let kiosk_ext = ext(self);
+        // Moved `OwnerToken` serves as proof of ownership
```
nmboavida, line 738: The `OwnerToken` is supposed to be used only for discoverability and not for proof of ownership given that the Kiosk's ownership is untransferable. I believe `assert_owner_address(self, sender(ctx));` should be added back here

nmboavida (changes requested):

PR #481, opened 2023-06-07
Fix dangling locks

nmboavida:


tomfurrier (approved):

contracts/kiosk/sources/kiosk/ob_kiosk.move
```diff
@@ -531,6 +534,9 @@ module ob_kiosk::ob_kiosk {
         ctx: &mut TxContext,
     ): (T, WithdrawRequest<T>) {
         assert_version_and_upgrade(ext(self));
+        // Delegated withdraws need to be settled via `withdraw_nft`
```
Suficio, line 537: Should you also add this assert to `withdraw_nft`?

Suficio (changes requested):

contracts/kiosk/sources/kiosk/ob_kiosk.move
nmboavida: The withdraw_nft requires a prior authorisation, which itself asserts if the NFT is being exclusively locked

